PR #53, opened 2015-02-09
fixing NameError in bugsnag.notify exception when bugsnag.log is called

rseaman:
Error squashes all other errors because bugsnag is not defined as a global here. log() is available after __init__.py fires by the time notify would be called.


ConradIrwin:
Thanks for this!

PR #67, opened 2016-01-04
Add sync mode

tume:
Adds new configuration parameter async (default True). When set to False will block until request is done.

Needed this behavior for integrating rq with bugsnag which doesn't seem to work very well with async transports.


kattrali:
Thanks, @tume, for the contribution. The code looks good. I'll look this over from an API viewpoint to ensure its consistent with other notifiers with similar functionality, and merge with a few documentation updates for the change.

kattrali:
Merged via #71 with documentation updates. Thanks again!

PR #68, opened 2016-01-12
Remove duplicate installation instructions in the README

kylef:
This makes it a little simpler, shorter and more readable.


kattrali:
Looks good, the single-bulleted list can just be converted to paragraphs and then :shipit:

PR #70, opened 2016-01-13
Derive Django example views from imported list instead of strings

kattrali:
Fixes some deprecation warnings regarding the use of a list of strings for view names


kylef:
:+1:

PR #71, opened 2016-01-13
Add an option to run synchronously

kattrali:
Merges #67, with documentation updates and a slight tweak for the property name to better match the others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,12 +210,12 @@ logger.addHandler(BugsnagHandler())
 *extra_fields*
 
 The BugsnagHandler accepts a special keyword argument to its `__init__()`
-function: 'extra_fields'.  This is optional and may be a dictionary of 
+function: 'extra_fields'.  This is optional and may be a dictionary of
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -282,6 +282,15 @@ Your Bugsnag API key (required).
 bugsnag.configure(api_key = "YOUR_API_KEY_HERE")
 ```
 
+### asynchronous
+
+By default, requests are sent asynchronously to Bugsnag. If you would like to
+block until the request is done, you can set `asynchronous`:
+
+```python
+bugsnag.configure(asynchronous = False)
```
kylef: No spaces around `=` here as per pep8:

> Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.
> Yes:
> ```python
> def complex(real, imag=0.0):
>     return magic(r=real, i=imag)
> ```
> No:
> ```python
> def complex(real, imag = 0.0):
>     return magic(r = real, i = image)
> ```

PR #74, opened 2016-01-14
Improve distribution

kattrali:
* Ensure secure PyPI upload with [twine](https://github.com/pypa/twine#why-should-i-use-this)
* Support [wheel distribution](http://pythonwheels.com) in addition to eggs


kylef:
:+1:

PR #75, opened 2016-01-15
Group tests in a more standard structure, add code style linting

kattrali:
* Reformat tests as `unittest` cases
* Split up sanitization tests to more easily check for breakage
* Simplify string sanitization
* Lint test code

### DIFF
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -2,8 +2,6 @@ import os
 import socket
 import unittest
 
-from nose.tools import eq_
-
 from bugsnag.configuration import Configuration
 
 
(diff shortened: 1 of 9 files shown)


tests/test_configuration.py
```diff
@@ -1,67 +1,69 @@
...
-    os.environ["DYNO"] = "YES"
-    c = Configuration()
-    assert(c.hostname == None)
+class TestConfiguration(unittest.TestCase):
+
+    def test_get_endpoint(self):
+        # Test default endpoint with ssl
+        c = Configuration()
+        c.use_ssl = True
+        eq_(c.get_endpoint(), "https://notify.bugsnag.com")
```
kylef: Why not, `self.assertEqual`?

tests/test_configuration.py
```diff
@@ -1,67 +1,69 @@
...
+        c.use_ssl = False
+        eq_(c.get_endpoint(), "http://notify.bugsnag.com")
+
+        # Test custom endpoint
+        c = Configuration()
+        c.use_ssl = False
+        c.endpoint = "localhost:1234"
+        eq_(c.get_endpoint(), "http://localhost:1234")
+
+    def test_environment_defaults(self):
```
kylef: Perhaps could be better named:

```python
def test_reads_api_key_from_environ(self):
```

tests/test_wsgi.py
```diff
@@ -14,138 +16,140 @@ class SentinalError(RuntimeError):
...
+                        user={"id": "5", "email":
+                              "me@cirw.in", "name": "conrad"})
+                raise SentinalError("oops")
 
-    eq_(deliver.call_count, 1)
-    payload = deliver.call_args[0][0]
-    eq_(payload['events'][0]['user']['id'], '5')
+        app = TestApp(BugsnagMiddleware(CrashAfterSettingUserId))
 
+        assert_raises(SentinalError, lambda: app.get('/beans'))
```
kylef: I'd use the unittest way for this:

```python
with self.assertRaises(SentinalError):
    app.get('/beans')
```

tests/test_configuration.py
kattrali: There were a few uses of `eq_` already and it helped with line length, though I agree that `self.assertEqual` is more readable.

tests/test_configuration.py
kattrali: Agreed.

tests/test_wsgi.py
kattrali: Makes sense. Also just noticed the typo.

PR #76, opened 2016-01-15
Fix user attributes for custom authentication

### LINKED ISSUE #62: User not set correctly when using custom auth in django
In the line:

```
notification.set_user(id=request.user.username, email=request.user.email, name=name)
```

It assumes that the username field is username. In custom authentication this isn't guaranteed to exist. You should do:

```
notification.set_user(id=request.user.get_username(), email=request.user.email, name=name)
```

kattrali:
Updated the user attributes to use `get_username()` instead of the `username` property as well as `get_full_name()` instead of concatenating (possibly unused) `first_name` and `last_name`.

Fixes #62

### DIFF
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import division, print_function, absolute_import
 
+import six
 from django.conf import settings
 from django.core.urlresolvers import resolve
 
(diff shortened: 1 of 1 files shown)


bugsnag/django/__init__.py
```diff
@@ -22,9 +22,8 @@ def add_django_request_to_notification(notification):
 
     if hasattr(request, 'user') and request.user.is_authenticated():
         try:
-            name = " ".join([request.user.first_name or '',
-                             request.user.last_name or ''])
-            notification.set_user(id=request.user.username,
+            name = request.user.get_full_name()
+            notification.set_user(id=request.user.get_username(),
                                   email=request.user.email, name=name)
```
kylef: `request.user.email` might also not exist, or it might not be a string.

Example that I might build a custom user object that makes the email a foreign relation to an `Email` model.

See [Specifying a custom User model](https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#specifying-a-custom-user-model)

kylef: Also, according to this documentation, the `USERNAME_FIELD` might be a foreign key as per:

> USERNAME_FIELD now supports ForeignKeys. Since there is no way to pass model instances during the createsuperuser prompt, expect the user to enter the value of to_field value (the primary_key by default) of an existing instance.

and:

> `get_username():`
>     Returns the value of the field nominated by USERNAME_FIELD.

Which could mean the username is also not a string, unsure if this causes any problems, just wanted to point out that it might not be a string.

kattrali: Hmm, both are good points. Patching.

PR #77, opened 2016-01-15
Log exception stack traces in middleware

### LINKED ISSUE #61: [py] Insufficient logging of exceptions in middleware
Currently, when an exception occurs in ``` MiddlewareStack.run(...) ``` only the exception message is exposed for troubleshooting.  Often, this is insufficient to resolve the underlying problem.

Example:
``` 
** [Bugsnag] Error in exception middleware: maximum recursion depth exceeded while calling a Python object
** [Bugsnag] Notifying https://notify.bugsnag.com of exception
```

The current run implementation in ``` bugsnag-python/bugsnag/middleware.py ``` is:
``` 
   def run(self, notification, callback):
        """
        Run all the middleware in order, then call the callback.
        """

        # the last step in the notification stack is to call the callback.
        # we also do this inside the exception handler, so need to ensure that
        # the callback is only called once.
        def finish(notification):
            if not hasattr(finish, 'called'):
                finish.called = True
                callback()

        to_call = finish
        for middleware in reversed(self.stack):
            to_call = middleware(to_call)

        try:
            to_call(notification)
        except Exception as exc:
            bugsnag.log("Error in exception middleware: %s\n%s" % exc)
            # still notify if middleware crashes before notification
            finish(notification)

``` 

My proposed solution would be:

```
   def run(self, notification, callback):
        """
        Run all the middleware in order, then call the callback.
        """

        # the last step in the notification stack is to call the callback.
        # we also do this inside the exception handler, so need to ensure that
        # the callback is only called once.
        def finish(notification):
            if not hasattr(finish, 'called'):
                finish.called = True
                callback()

        to_call = finish
        for middleware in reversed(self.stack):
            to_call = middleware(to_call)

        try:
            to_call(notification)
        except Exception as exc:
>>>>        stack_trace = traceback.format_exc()
>>>>        bugsnag.log("Error in exception middleware: %s\n%s" % (exc, stack_trace))
            # still notify if middleware crashes before notification
            finish(notification)

```

... which would require ``` import traceback ```

kattrali:
Update middleware logging to include the stack trace in addition to the name and description of the exception. Alternately, we can fix this by reraising exception, but I have no context around the current behavior. Any thoughts?

Fixes #61

### DIFF
--- a/bugsnag/middleware.py
+++ b/bugsnag/middleware.py
@@ -1,3 +1,5 @@
+import traceback
+
 import bugsnag
 
 
@@ -120,6 +122,7 @@ class MiddlewareStack(object):
         try:
             to_call(notification)
         except Exception as exc:
-            bugsnag.log("Error in exception middleware: %s" % exc)
+            template = "Error in exception middleware: %s\n%s"
+            bugsnag.log(template % (exc, traceback.format_exc()))
             # still notify if middleware crashes before notification
             finish(notification)


snmaynard:
I think we just want to log the exception, but ultimately still send the exception. Reraising wouldnt get notified to bugsnag so that might be confusing. Notifying bugsnag would also trigger the exception again, unless we somehow skipped middleware when notifying the exception.

I think logging is the best compromise - thoughts?

kattrali:
Yeah, I think that was the context I needed, thanks.

kattrali:
Wait, so is that a yes or no to including the stacktrace in the log?

snmaynard:
Stacktrace in the log seems like a great idea to me!

PR #78, opened 2016-01-15
Remove cookies from request payloads

### LINKED ISSUE #30: Move cookies into their own tab
Our ruby notifier has cookies in their own tab, whereas in python we nest them under the request tab. They are much more readable in their own tab, so I propose we move cookies to their own tab in python too.

kattrali:
Fixes #30, yay security


snmaynard:
LGTM!

PR #86, opened 2016-08-10
Use the requests library if it's installed

GrahamCampbell:


bugsnag/notification.py
```diff
@@ -20,27 +20,49 @@
...
+            if proxy_host:
+                proxy = {
+                    'https': proxy_host,
+                    'http': proxy_host
+                }
+                r = requests.post(url, data=payload, proxies=proxy)
+            else:
+                r = requests.post(url, data=payload)
+
+            status = r.status_code;
```
kylef: Semi-colon in Python? 😬

bugsnag/notification.py
```diff
@@ -20,27 +20,49 @@
...
+
+            if status != 200:
```
kylef: I'd change the hard-coded 200 value to `requests.codes.ok`

bugsnag/notification.py
GrahamCampbell: haha, thanks!

bugsnag/notification.py
GrahamCampbell: Ok, sure. :)

gkze:
LGTM. Maybe add tests? CC @kattrali

GrahamCampbell:
🚀

PR #87, opened 2016-08-16
Improve payload sanitization

kattrali:
This changeset increases the amount of payload sanitization while consolidating the sanitization and encoding logic into a `JSONEncoder`.

## Changes

* Handle recursive fields in payload
* Handle coercion to unicode strings in non-metadata fields
* Remove imports of `simplejson` as it is not applicable on Python 2.6+
* Improve test coverage

### DIFF
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -3,93 +3,91 @@ from __future__ import division, print_function, absolute_import
 import inspect
 import traceback
 import six
+from json import JSONEncoder
 
 import bugsnag
-
-try:
-    import json
-except ImportError:
-    import simplejson as json
 
 
 MAX_PAYLOAD_LENGTH = 128 * 1024
 MAX_STRING_LENGTH = 1024
 
 
-def sanitize_object(obj, **kwargs):
-    ignored = kwargs.get("ignored", [])
-    kwargs["ignored"] = ignored
+class SanitizingJSONEncoder(JSONEncoder):
+    """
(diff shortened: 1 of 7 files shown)


bugsnag/utils.py
```diff
@@ -1,78 +1,92 @@
...
-                if clean_obj is not None:
-                    clean_dict[k] = clean_obj
-
-        return clean_dict
-    elif any(isinstance(obj, t) for t in (list, set, tuple)):
-        return [sanitize_object(x, **kwargs) for x in obj]
-    elif any(isinstance(obj, t) for t in (bool, float, int)):
-        return obj
-    else:
+class JSONEncoder(json.JSONEncoder):
```
gkze: If these 2 classes were `import`ed into the same namespace they would caused a conflict unless they were imported with `as` having different names. I would call this `SanitizingJSONEncoder`

I'm being nitpicky so feel free to ignore 😄

tests/test_flask.py
```diff
@@ -93,8 +93,9 @@ def hello():
             raise SentinelError("oops")
 
         handle_exceptions(app)
-        app.test_client().get('/hello')
-        app.test_client().get('/hello')
+        with app.test_client() as client:
```
gkze: context managers FTW

gkze:
LGTM overall!

bugsnag/utils.py
kattrali: Yeah, that sounds like a good idea. Patching.

PR #97, opened 2016-08-22
No longer register a global proxy opener

kylef:
Bugsnag should not affect the global urllib2 state and only register the proxy for the specific request.

This commit also adds test coverage for the proxy itself.


kattrali:
Thank you, @kylef!

PR #102, opened 2016-09-08
Introduce a new client based architecture

kylef:
This pull request introduces a new architecture to bugsnag python where you can create multiple clients which each themselves have their own configuration.

```python
client = bugsnag.Client()
client.notify(Exception('hallo its me'))
```

I've also introduced a `capture` method to the client which allows you to capture exceptions in a with context against the client.

```python
with client.capture():
    raise Exception('captured exception')
```

This can also be used as a decorator against other functions.

```python
@client.capture()
def my_function():
    raise Exception('captured exception')
```

Please see the changes to CHANGELOG for further information with these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,90 @@
 Changelog
 =========
 
+## 2.6.0b1 (TBA)
+
+### Enhancements
+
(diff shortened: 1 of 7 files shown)


kattrali:
🚀

PR #105, opened 2016-09-13
Deprecate use_ssl and get_endpoint

kattrali:
Endpoint can and should include the protocol, without the indirection


att14:
Since the default is `True` you get this warning even if you aren't setting this option, which doesn't seem like the correct behavior.

PR #108, opened 2016-09-30
Have BugsnagMiddleware for Django inherit from Django MiddlewareMixin…

jonnypickett:
… for 1.10 compatibility


bugsnag/django/middleware.py
```diff
@@ -1,12 +1,15 @@
 from __future__ import division, print_function, absolute_import
 
+from django.utils.deprecation import MiddlewareMixin
```
kattrali: To make this backwards compatible, this call should be wrapped to ensure no import error occurs and there is a fallback base class. Something like:

```python
try:
    from django.utils.deprecation import MiddlewareMixin
except ImportError:
    MiddlewareMixin = object
```

jonnypickett: Good catch. But actually now that you point this out, passing a parameter to `object.__init__()` throws `TypeError: object.__init__() takes no parameters`, which is what will happen for Django projects < 1.10 with passing `get_response` to the parent `__init__` method.

jonnypickett: Great. Glad I could help.

bugsnag/django/utlis.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import print_function, absolute_import
```
kattrali: Oh, just noticed the typo in the filename

PR #119, opened 2017-06-07
FIX - Passing compiled modules to traceback_exclude_modules

kobim:
`Configure.traceback_exclude_modules` receives a list of modules, and extracts the `__file__` of each, allowing to skip few modules from appearing in the stacktrace.
In Python 2.x, when a compiled module is loaded, it's `__file__` property displays the compiled `.pyc` path, while the running code will appear in the stacktrace with a `.py` suffix.

This fix replaces the _.pyc_ suffix with a _.py_, hence enabling the `traceback_exclude_modules` to work as it should.

I also added tests reproducing this scenario, and also tweaked _.travis.yml_ to support the new image they deployed recently (https://blog.travis-ci.com/2017-05-04-precise-image-updates).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,59 @@
 language: python
 
-env:
-  - TOXENV=py26-test
-  - TOXENV=py27-test
-  - TOXENV=py33-test
-  - TOXENV=py34-test
-  - TOXENV=py35-test
-  - TOXENV=py26-requests-test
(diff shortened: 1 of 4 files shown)


kattrali:
Thank you for the very thorough report and patch, @kobim. Looking it over now.

kattrali:
Looks great! Thanks for updating the CI config as well. 🎉

PR #122, opened 2017-08-15
Added flask example

Cawllec:
Added example of how to use the library in a Flask Python app covering issues on-par with currently existing examples:
- Config
- Crashes
- Crashes with additional data attached
- Notifying of handled errors
- Notifying with additional diagnostic data
- Modifying the context of a notification
- Modifying the severity of a notification

Is it worth considering additional coverage that could then be included in the other example apps?


example/flask/README.md
```diff
@@ -0,0 +1,55 @@
+# Bugsnag Flask demo
+
+This Flask application demonstrates how to use Bugsnag with the Flask web
+framework for Python. Before testing it, open up the `server.py`
+file and configure your API key.
+
+```
```
kylef: You can use the following to add syntax highlighting to the rendered code-block:

    ```python

kattrali:
Thanks for taking a look, @kylef!

Cawllec:
I've updated the code and quote styling to be more pythonic, and added the default index as suggested

example/flask/requirements.txt
```diff
@@ -8,3 +8,27 @@ MarkupSafe==1.0
 six==1.10.0
 WebOb==1.7.3
 Werkzeug==0.12.2
+blinker==1.4
+bugsnag==3.1.1
```
Cawllec: Good catch, have that cleared up now

PR #124, opened 2017-08-17
Cawllec/django+celery example

Cawllec:
Added example of how to use the library in a Django app utilising Celery alongside as a task runner covering issues on-par with currently existing examples:

Config
Crashes
Crashes with additional data attached
Notifying of handled errors
Notifying with additional diagnostic data
Modifying the context of a notification
Modifying the severity of a notification


kattrali:
🎉

PR #126, opened 2017-09-13
Adding Handled/Unhandled payload properties

Cawllec:
- Adds `defaultSeverity`, `unhandled`, and `severityReasons` payload properties to support upcoming handled/unhandled functionality
- Above properties added through **options parameters
- System error handler, log level handlers, and middleware error handlers set up with severity reasons
_references [PLAT_209](https://bugsnag.atlassian.net/browse/PLAT-209)_

### DIFF
--- a/bugsnag/celery/__init__.py
+++ b/bugsnag/celery/__init__.py
@@ -14,7 +14,14 @@ def failure_handler(sender, task_id, exception, args, kwargs, traceback, einfo,
 
     bugsnag.auto_notify(exception, traceback=traceback,
                         context=sender.name,
-                        extra_data=task)
+                        extra_data=task,
+                        unhandled=True,
+                        severity_reason={
+                            'type':'middleware_handler',
+                            'attributes': {
+                                'name': 'celery'
+                            }
+                        })
 
 
 def connect_failure_handler():
--- a/bugsnag/celery/__init__.py
+++ b/bugsnag/celery/__init__.py
@@ -16,9 +16,9 @@ def failure_handler(sender, task_id, exception, args, kwargs, traceback, einfo,
                         context=sender.name,
                         extra_data=task,
                         severity_reason={
-                            'type': 'middleware_handler',
+                            'type': 'unhandledExceptionMiddleware',
                             'attributes': {
-                                'name': 'celery'
+                                'framework': 'Celery'
                             }
                         })
 
--- a/bugsnag/celery/__init__.py
+++ b/bugsnag/celery/__init__.py
@@ -17,7 +17,7 @@ def failure_handler(sender, task_id, exception, args, kwargs, traceback, einfo,
                         extra_data=task,
                         unhandled=True,
                         severity_reason={
-                            'type':'middleware_handler',
+                            'type': 'middleware_handler',
                             'attributes': {
                                 'name': 'celery'
                             }
--- a/bugsnag/celery/__init__.py
+++ b/bugsnag/celery/__init__.py
@@ -15,7 +15,6 @@ def failure_handler(sender, task_id, exception, args, kwargs, traceback, einfo,
     bugsnag.auto_notify(exception, traceback=traceback,
                         context=sender.name,
                         extra_data=task,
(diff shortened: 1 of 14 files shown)


bugsnag/wsgi/middleware.py
```diff
@@ -28,13 +28,23 @@ class WrappedWSGIApp(object):
 
     def __init__(self, application, environ, start_response):
         self.environ = environ
+        self.severity_reason = {
```
snmaynard: This should be a constant rather than a member on the middleware

bugsnag/client.py
```diff
@@ -117,12 +120,24 @@ def deliver(self, notification):  # type: (Notification) -> None
...
+        initial_unhandled = notification.unhandled
+        initial_severity_reason = notification.severity_reason
+
         def send_payload():
             if notification.api_key is None:
                 bugsnag.logger.warning(
                     "No API key configured, couldn't notify")
                 return
 
+            notification.default_severity = initial_default_severity
```
snmaynard: I dont think we need to reset any changes to these variables in here. If people change them then they are in expert mode and thats fine.

bugsnag/celery/__init__.py
```diff
@@ -14,7 +14,14 @@ def failure_handler(sender, task_id, exception, args, kwargs, traceback, einfo,
 
     bugsnag.auto_notify(exception, traceback=traceback,
                         context=sender.name,
-                        extra_data=task)
+                        extra_data=task,
+                        unhandled=True,
```
snmaynard: is this not implied because auto notify?

bugsnag/legacy.py
```diff
@@ -47,6 +47,9 @@ def notify(exception, **options):
     """
     Notify bugsnag of an exception.
     """
+    if 'severity' in options:
```
snmaynard: this will trigger for all auto notify calls which is incorrect, you still want defaultSeverity to be true in those cases

Cawllec:
I've updated the notify default-severity check, that shouldn't automatically set default_severity to false when unhandled calls to it are made.  I've taken out the overriding of severity_reason and unhandled, but this is a pattern in some of the other notifiers (whether through making them protected properties or by overriding them if necessary) which may need discussion

bugsnag/client.py
```diff
@@ -117,12 +122,22 @@ def deliver(self, notification):  # type: (Notification) -> None
...
                 bugsnag.logger.warning(
                     "No API key configured, couldn't notify")
                 return
 
+            if initial_severity != notification.severity:
+                notification.severity_reason = {
+                    'type': 'userCallbackSetSeverity'
+                }
+            else:
+                notification.severity_reason = initial_reason
```
snmaynard: Why do we do this here?

bugsnag/client.py
```diff
@@ -150,6 +165,8 @@ class ClientContext(object):
     def __init__(self, client, exception_types=None, **options):
         self.client = client
         self.options = options
+        if 'severity' in options:
+            options['severity_reason'] = dict(type='contextSpecifiedSeverity')
```
snmaynard: Should this be `userContextSetSeverity`?

bugsnag/handlers.py
```diff
@@ -37,7 +37,16 @@ def emit(self, record):
             if path in record.pathname:
                 return
 
-        options = {'meta_data': {}}
+        options = {
+            'meta_data': {},
+            'unhandled': False,
+            'severity_reason': {
+                'type': 'logs',
```
snmaynard: This should be log

bugsnag/client.py
Cawllec: This enforces that users cannot overwrite the severity_reason in callbacks or middleware.  I think we previously said it wouldn't matter if someone chose to do so, but we've added it as one of our testing parameters and are enforcing the same thing across notifiers.

bugsnag/client.py
Cawllec: This one seems to already have been fixed

PR #130, opened 2017-11-01
urlresolvers removed from django-2.0

overplumbum:
`django.core.urlresolvers` has been removed from django-2.0 causing bugsnag-python to crash on initialization.

I've added new naming scheme fallback.
Also configured tox to run tests against django-2.0.

### DIFF
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -2,7 +2,11 @@ from __future__ import division, print_function, absolute_import
 
 import six
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


overplumbum:
tests fail only py35lint check in unaffected files (clearly should be in a separate PR)

Besides I guess it's a good moment to start testing against django-2.0

Cawllec:
Hi overplumbum, thanks for the contribution.  This looks good to me, I'll fix the issues with the lint check in the other files and look at getting this included.

Issue #132, opened 2017-11-08
Support X-Forwared-For header for client IP


wosc:
When running an application behind (possibly several) HTTP reverse proxies (nginx, haproxy, etc.) bugsnag should still report the original client IP address as the "user" information, and not the one of the most recent proxy (more often than not: localhost). This is already supported by `webob`, so the necessary change would be:

```diff
diff --git a/bugsnag/wsgi/middleware.py b/bugsnag/wsgi/middleware.py
index 7ecf160..d28f7f1 100644
--- a/bugsnag/wsgi/middleware.py
+++ b/bugsnag/wsgi/middleware.py
@@ -12,7 +12,7 @@ def add_wsgi_request_data_to_notification(notification):
     request = Request(environ)
 
     notification.context = "%s %s" % (request.method, request_path(environ))
-    notification.set_user(id=request.remote_addr)
+    notification.set_user(id=request.client_addr)
     notification.add_tab("request", {
         "url": request.path_url,
         "headers": dict(request.headers),
```

PR #133, opened 2017-11-08
Support X-Forwarded-For for client IPs

wosc:
As described in #132


kylef:
This would make it relatively easy for malicious users to provide incorrect IP via the `X-Forwarded-For` header when the WSGI server is not behind a proxy server or a different configured header (for example `X-Real-IP`). Perhaps this should be opt-in vs the default behaviour.

PR #138, opened 2018-01-26
Logger example app for Python

tremlab:
A very simple, terminal-based, framework-agnostic example of how Bugsnag can collect all logs & exceptions through your existing python logger, rather than explicitly calling bugsnag.notify().


example/logger/README.md
```diff
@@ -0,0 +1,34 @@
...
+    git clone git@github.com:bugsnag/bugsnag-python.git
+    cd bugsnag-python/example/logger
+    ```
+
+1. Install dependencies
+    ```shell
+    pip install -r requirements.txt
+    ```
+
+1. Before testing it, open up the `server.py` file and configure your API key.
```
Cawllec: This is instruction should be for the `bugsnag_logger.py` file

example/logger/bugsnag_logger.py
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    notification.add_tab(
+        'company', {
+            'name': 'Stark Industries'
+        }
+    )
+
+    # The callback will evaluate all exceptions, but in this example only 'SomethingBad' errors will have the below data added to their error reports.
+    if notification.exception.message == 'SomethingBad':
```
Cawllec: This is causing an exception to be thrown for me here, `AttributeError: 'MyCustomException' object has no attribute 'message'`

example/logger/server.py
```diff
@@ -0,0 +1,54 @@
+from bugsnag_logger import logger
+# This file demonstrates how errors will be reported to bugsnag through your logger, even without any explicit reference to bugsnag in your python code.
+
+print """
```
Cawllec: All `print` statements need to be bracketed for Python 3+ compatability

example/logger/server.py
```diff
@@ -0,0 +1,54 @@
...
+
+def handle_zero_div():
+    """Deliberately triggers a handled exception, and reports it to Bugsnag.
+    """
+    try:
+        x = 1/0
+    except Exception as e:
+        logger.warn(e)
+
+    print 'The app hasn\'t crashed, but check bugsnag.com to view notifications'
```
Cawllec: As mentioned above

example/logger/server.py
```diff
@@ -0,0 +1,54 @@
...
+
+def log_error():
+    """Simply logs an error, which will also be sent to your Bugsnag dashboard.
+    """
+    logger.error('I forgot my lunch at home')
+    print 'Check bugsnag.com to view this log report.'
```
Cawllec: As mentioned above

Cawllec:
This example has some overlap with the plain example PR, would it be feasible to merge them?

PR #139, opened 2018-01-30
Django example app - expanding & updating

tremlab:
Added more detailed configuration examples.  Updated to Django2 with Python3. (not backwards compatible, due to Django's dependency.)


tremlab:
a minor update to the notifier is needed to make this app compatible with the latest version of Django, but let me know any comments on the app itself.

example/django/README.md
```diff
@@ -1,69 +1,36 @@
...
 
-1. [crash and use callbacks](/crash_with_callback)
-<br/>
-Before crashing, the library would append the Diagnostics tab with some
-predefined information, attached by means of a callback.
-Pressing this link would lead to an empty page, which is normal.
-See `demo/views.py, #crash_with_callback`
+1. Run the application. (Make sure to use any 3+ version of python.)
+    ```shell
+    python3.6 manage.py runserver
```
Cawllec: I don't know that we want to go this granular with the python command, I think `python` would be better than `python3.6`

example/django/README.md
```diff
@@ -1,69 +1,36 @@
...
+    pip3 install -r requirements.txt
+    ```
 
-1. [crash](/crash)
-<br/>
-Crashes the library and sends a notification about the nature of the crash.
-Basically, almost any unhandled exception sends a notification to Bugsnag.
-Pressing this link would lead to an empty page, which is normal.
-See `demo/views.py, #crash`
+1. Before testing it, open up the `settings.py`
```
Cawllec: `settings.py` is found in `bugsnag_demo`, would it be worth pointing directly to the file i.e. `bugsnag_demo/settings.py`

example/django/bugsnag_demo/settings.py
```diff
@@ -38,7 +38,8 @@
     'django.contrib.staticfiles',
 )
 
-MIDDLEWARE_CLASSES = (
+MIDDLEWARE = (
+    # make sure to add Bugsnag to your middleware.
```
Cawllec: While this step is mentioned in the getting started docs, I think it can be overlooked and it may deserve a step in the readme.

example/django/index.html
```diff
@@ -0,0 +1,43 @@
...
+        <a href="/crashcallback">
+          <button type="button" id="indexError">UNHANDLED WITH METADATA</button>
+        </a>
+        <br>
+        <a href="/handled">
+          <button type="button" id="typeError">HANDLED ERROR</button>
+        </a>
+        <br>
+        <a href="/notifywithcontext">
+          <button type="button" id="typeError">MANUAL NOTIFICATION</button>
```
Cawllec: `MANUAL NOTIFICATION` doesn't give a great indication of what the button is actually doing.  It would be good to differentiate between that and `HANDLED ERROR`.

example/django/index.html
```diff
@@ -0,0 +1,43 @@
...
+    <h3>This page is a basic example of how to include bugsnag-python in a django app.
+    </h3>
+  </div>
+
+    <br>
+  <h3>Click below to send errors to your Bugsnag dashboard.</h3>
+  <br>
+  <br>
+    <div id="python">
+        <a href="crash/">
```
Cawllec: Stray underlines can be fixed with `style="text-decoration:none;"`

example/django/README.md
```diff
@@ -1,69 +1,36 @@
...
-framework for Python. Before testing it, open up the `bugsnag_demo/settings.py`
-file and configure your API key (see the very bottom of the file).
+This Django application demonstrates how to use Bugsnag with the Django web framework for Python.
 
-```
-BUGSNAG = {
-    "api_key": "066f1ad3590596f9aacd601ea89af845"
-}
-```
+Please note that while bugsnag-python is compatible with previous Python 2 and Django 1 versions -- however, this particular example app will only work with Python 3+ and Django 2+, to showcase the most current configurations. See [our documentation](https://docs.bugsnag.com/platforms/python/django/) for further details.
```
Cawllec: I think this is a good idea, but I also think it'd be good to maintain a Django 1.11 example, maybe in a `django_legacy` folder?

kattrali: If anything, I'd call it `django1.11` for clarity

tremlab: sounds good!  I had built one, but then worried it would be too redundant. I'll rename it and submit in a separate PR.

tremlab: added django 1.10 in this same PR.

tremlab:
thanks for the detailed notes!  I believe my latest commits address all of them.

Cawllec:
Looks good!  I'd like to change the example directory name to `django2` and we need to update the remote before merging, but otherwise seems fine. I need to fix the build as well as the tests require the correct example, so I'll take it from here.

Cawllec:
This is probably going to require refactoring the tests from using the examples to using proper test-fixtures, which is going to take a little longer.  This is still active though!

Cawllec:
Merged as a part of [this PR](https://github.com/bugsnag/bugsnag-python/pull/152)

PR #140, opened 2018-01-31
version of our example app compatible with Django 1.10

tremlab:
This version of the Django example app will work with Python <3, and Django <2.


Cawllec:
Is this intended to sit alongside `django` as examples for django 1 & 2?  If so, can we have this example named `django` and the django 2 example `django2`?

tremlab:
After some discussion, we only need one Django example - v2 / python 3.  Will close this one. thanks!

PR #141, opened 2018-02-02
Allow the 'exception' option to override the logged exception in a handler callback

Cawllec:
-Tests whether `exception` is set in options before notifying of a log entry
- If it is and is an instance of an `Exception`, notify using that exception
- If it is and it isn't an instance of an `Exception` add it to metadata and remove original key to avoid name conflicts.

### DIFF
--- a/bugsnag/handlers.py
+++ b/bugsnag/handlers.py
@@ -56,6 +56,13 @@ class BugsnagHandler(logging.Handler, object):
 
         client = self.client or bugsnag.legacy.default_client
 
+        if 'exception' in options:
+            if isinstance(options['exception'], Exception):
+                client.notify(**options)
+                return
+            else:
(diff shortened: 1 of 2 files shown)


bugsnag/handlers.py
```diff
@@ -56,6 +56,13 @@ def emit(self, record):
 
         client = self.client or bugsnag.legacy.default_client
 
+        if 'exception' in options:
+            if isinstance(options['exception'], Exception):
+                client.notify(**options)
+                return
+            else:
+                options['meta_data']['exception'] = options.pop('exception')
```
kattrali: Potentially want a check here to ensure that none of the callbacks changed the type of the `meta_data` object and it can be subscripted.

Cawllec: I've updated to drop the `options['exception']` in the case it's not a dict anymore.

PR #143, opened 2018-02-19
Ensured is_authenticated is used correctly

Cawllec:
Fixes warnings that `is_authenticated` is not callable.

This has been tested manually on Django 1.11 & 2.0.

### DIFF
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -26,14 +26,19 @@ def add_django_request_to_notification(notification):
             notification.context = "%s %s" % (request.method,
                                               request.path_info)
 
-    if hasattr(request, 'user') and request.user.is_authenticated():
-        try:
-            name = request.user.get_full_name()
-            email = getattr(request.user, 'email', None)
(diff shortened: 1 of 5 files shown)


bugsnag/django/__init__.py
```diff
@@ -26,14 +26,19 @@ def add_django_request_to_notification(notification):
...
-    if hasattr(request, 'user') and request.user.is_authenticated():
-        try:
-            name = request.user.get_full_name()
-            email = getattr(request.user, 'email', None)
-            username = six.text_type(request.user.get_username())
-            notification.set_user(id=username, email=email, name=name)
-        except Exception:
-            bugsnag.logger.exception('Could not get user data')
+    if hasattr(request, 'user'):
+        if hasattr(request.user.is_authenticated, '__call__'):
```
kylef: There is a `callable` method, maybe better to use that instead of checking for `__call__` method.

```python
if callable(request.user.is_authenticated):
```

Cawllec: You're correct, I was following outdated information

PR #148, opened 2018-04-03
Speed up SanitizingJSONEncoder on large objects

jonls:
Fixes an issue where trying to attach a large payload to a notification would take a long time. Changes the `SanitizingJSONEncoder` to use `set` instead of `list` for faster lookup of ignored objects.


martin308:
@jonls thanks for the contribution 🙏 this is definitely a big speed up for large object! 

I have added some tests to check the speed of json encoding and made the methods backwards compatible if people are providing a list as the parameter here. I will push up a PR with the changes shortly

PR #149, opened 2018-04-05
Speed up SanitizingJSONEncoder on large objects w/ tests

martin308:
Built on the great work of @jonls. 

This adds some tests and makes sure that the methods are backwards compatible.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -155,3 +155,23 @@ encoder.encode(data)
         """
         time = timeit.timeit(stmt=stmt, setup=setup, number=1000)
         self.assertTrue(time < 4)
+
+    def test_filter_string_values_list_handling(self):
+        """
+        Test that filter_string_values can accept a list for the ignored
+        parameter for backwards compatibility
+        """
+        data = {}
+        encoder = SanitizingJSONEncoder()
+        # no assert as we are just expecting this not to throw
+        encoder.filter_string_values(data, ['password'])
+
+    def test_sanitize_list_handling(self):
+        """
+        Test that _sanitize can accept a list for the ignored parameter for
+        backwards compatibility
+        """
+        data = {}
+        encoder = SanitizingJSONEncoder()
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -136,3 +138,46 @@ def test_encoding_oversized_recursive(self):
...
+        stmt = """\
+encoder.encode(data)
+        """
+        time = timeit.timeit(stmt=stmt, setup=setup, number=1000)
+        maximum_time = 4
+        if sys.version_info[0:2] <= (2, 6):
+            # json encoding is very slow on python 2.6 so we need to increase
+            # the allowable time when running on it
+            maximum_time = 18
+        self.assertTrue(time < maximum_time)
```
kattrali: I'd alter the error message to print the values here:

```diff
- self.assertTrue(time < maximum_time)
+ self.assertTrue(time < maximum_time,
+                 "Encoding required {}s (expected {}s)".format(
+                     time, maximum_time
+                 ))
```

tests/test_utils.py
```diff
@@ -136,3 +138,49 @@ def test_encoding_oversized_recursive(self):
...
+        stmt = """\
+encoder.encode(data)
+        """
+        time = timeit.timeit(stmt=stmt, setup=setup, number=1000)
+        maximum_time = 6
+        if sys.version_info[0:2] <= (2, 6):
+            # json encoding is very slow on python 2.6 so we need to increase
+            # the allowable time when running on it
+            maximum_time = 18
+        self.assertTrue(time < maximum_time,
```
Cawllec: I don't like this as a pass/fail metric for tests, as it's too dependent on the performance of the CI tools.

kattrali: Its dependent on the tools, but also a small canary if something goes terribly wrong under the lax requirements of this test. I'd prefer that we have something for now and look into something more thorough in the future, like a formal perf suite.

PR #155, opened 2018-05-18
Remove 3.3 ci

Cawllec:
## Goal

Python 3.3 is deprecated, and the dropping of support from `wheel` and `setuputils` are causing tests to fail.

This PR removes 3.3 from the tests and tox environment, and adds the newest version of Python, 3.6.  It also moves the tests targeting 3.5 to 3.6, such as linting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,13 +28,6 @@ matrix:
   - python: 2.7
     env: TOXENV=py27-django110-migrate1-django1
 
-  - python: 3.3
-    env: TOXENV=py33-test
-  - python: 3.3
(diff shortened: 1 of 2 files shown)


fractalwrench:
I think this could be merged into master if it makes sense, as it only touches build files.

browniebroke:
Are you planning to merge this? I'd like to contribute and add newer versions of Django but this has been sitting for 2+ months leaving the build broken.

fractalwrench:
Thanks for the reminder @browniebroke - the original PR author is currently on leave which is why this has been hanging around for a while. I'll merge it now.

PR #158, opened 2018-08-01
Add Django 1.11, 2.0 and 2.1 to the build matrix

browniebroke:
## Goal

Django 2.1 [was released today](https://www.djangoproject.com/weblog/2018/aug/01/django-21-released/), which means all versions of Django in the build matrix are currently no longer supported by the Django core team. This adds newer versions of Django to the build matrix. 

You might want to drop support for older versions <1.11 and bump the version of the library. Let me know if you want me to do so as part of this PR, I don't mind it.

## Design

n/a

## Tests

Running Python 2.7 tests under Django 1.11 but only Python 3.6 on Django 2.0 and above, as the project dropped support for Python 2.

## Review

  - [ ] Final review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ matrix:
     env: TOXENV=py27-django19-migrate1-django1
   - python: 2.7
     env: TOXENV=py27-django110-migrate1-django1
+  - python: 2.7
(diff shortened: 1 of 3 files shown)


fractalwrench:
Many thanks for this contribution @browniebroke!

PR #159, opened 2018-08-03
Fix for ignore_classes setting for django's 404

browniebroke:
## Goal

In my setup, by default, Bugsnag logged some of Http404 despite the default config telling otherwise.

Looking at the examples, it was not super clear that I had to set the value to the full exception path in the setting. Then I managed to fix my setup by updating the exception path which I noticed was slightly different.

- Updating the examples to be clearer by specifying the full exception path
- Updating the default config to expect Http404 with full path or shortcut path

### Changed

Default `ignore_classes` option was changed to cover `django.http.response.Http404`. The existing path is a valid import so I left it around.

## Review

  - [ ] Final review

### DIFF
--- a/bugsnag/configuration.py
+++ b/bugsnag/configuration.py
@@ -63,7 +63,11 @@ class Configuration(_BaseConfiguration):
         self.app_version = None
         self.params_filters = ["password", "password_confirmation", "cookie",
                                "authorization"]
-        self.ignore_classes = ["KeyboardInterrupt", "django.http.Http404"]
+        self.ignore_classes = [
+            "KeyboardInterrupt",
+            "django.http.Http404",
+            "django.http.response.Http404",
+        ]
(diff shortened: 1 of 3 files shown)


fractalwrench:
Thanks for the contribution @browniebroke - this seems like a reasonable changeset to me. @kattrali will be able to review this as well just to get a second pair of eyes.

The only thing I noticed is that the `ignore_classes` hasn't been updated for some of the example projects (flask, logger, plain).

browniebroke:
Yes, I agree it would probably make sense to update the other example projects to be consistent, but I'm only using this in a Django project. I don't know if there is a built-in Flask exception which would make sense to Flask users? Maybe the custom one should have a dummy package in front as well?

I'm open to suggestions to update all the examples.

fractalwrench:
Added a changelog for this in #160 - thanks again for the contribution!

Issue #162, opened 2018-08-31
Exception in SanitizingJSONEncoder when serializing invalid payload json


brockhaywood:
### Description
Under some unknown circumstances, the bugsnag middleware is crashing when serializing the data to be sent to bugnsag. This is particularly problematic because the original exception is then hidden so it's difficult to determine what the data was that caused the snag to throw an exception.

The exception is being thrown in the `SanitizingJSONEncoder.encode` function. 

### Environment

Library versions:

- python version: 2.7
- bugsnag-python version: 3.4.2

### Example code snippet

```python
from bugsnag.utils import SanitizingJSONEncoder
def foo():
   return '123'

bar = {foo: '456'}

enc = SanitizingJSONEncoder()
enc.encode(bar)
```
This is obviously a pretty contrived example but we are seeing this in our production codebase and are, at present, unable to determine where the key is being set to something other than a string. My guess is that our codebase is attempting to serialize some json that is invalid and this invalid json is bubbling up into the bugsnag reporting and causing an issue there, as well.

```
2018-08-31 21:32:04,591 - [bugsnag] WARNING - No API key configured, couldn't notify
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/bugsnag/utils.py", line 34, in encode
    payload = super(SanitizingJSONEncoder, self).encode(safe_obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be a string
```
I wonder if we can sanitize the keys prior to passing them through to the json encoder.

PR #163, opened 2018-08-31
Sanitizes the keys passed in the payload dictionary to the SanitizingJSONEncoder

brockhaywood:
…JSONEncoder

## Goal
In some circumstances, it may be possible to crash the SanitizingJSONEncoder by passing bad keys in the payload to the SanitizingJSONEncoder. This PR attempts to sanitize the keys to ensure they are all strings.

### Changed
Changed the `_sanitize_dict` function of the `SanitizingJSONEncoder` to ensure that all keys are strings. If a key cannot be converted to a string then it is dropped from the clean dictionary as this will crash with it.

## Tests

Added `test_json_encode_invalid_keys` to `TestUtils`

### Linked issues

Fixes https://github.com/bugsnag/bugsnag-python/issues/162

## Review

  - [ ] Final review


bugsnag/utils.py
```diff
@@ -118,7 +118,19 @@ def _sanitize_dict(self, obj, trim_strings, ignored):
...
+            if isinstance(key, basestring):
+                clean_dict[key] = clean_value
+            else:
+                try:
+                    clean_dict[str(key)] = clean_value
+                except:
+                    try:
+                        clean_dict[repr(key)] = clean_value
+                    except:
+                        pass
```
brockhaywood: If we can't sanitize the key then drop it from the clean_dict otherwise we know this will crash.

bugsnag/utils.py
```diff
@@ -118,7 +118,19 @@ def _sanitize_dict(self, obj, trim_strings, ignored):
 
         clean_dict = {}
         for key, value in six.iteritems(obj):
-            clean_dict[key] = self._sanitize(value, trim_strings, ignored)
+
+            clean_value = self._sanitize(value, trim_strings, ignored)
+
+            if isinstance(key, basestring):
```
Cawllec: Our library is used in both Python 2.7 and 3, as such `six.string_types` is necessary here for compatibility.

bugsnag/utils.py
```diff
@@ -118,7 +118,19 @@ def _sanitize_dict(self, obj, trim_strings, ignored):
...
+                clean_dict[key] = clean_value
+            else:
```
Cawllec: This may be cleaner split out into a separate function

bugsnag/utils.py
```diff
@@ -118,7 +118,19 @@ def _sanitize_dict(self, obj, trim_strings, ignored):
...
+                try:
+                    clean_dict[str(key)] = clean_value
+                except:
+                    try:
+                        clean_dict[repr(key)] = clean_value
```
Cawllec: We should be wary of using the `repr` function response as a key, it may produce a lot more information than was intended for use in this area.

bugsnag/utils.py
Cawllec: In the event an object couldn't be converted into a string using either `str` or `repr`, this could cause further problems as data would be being dropped with no indication.  If we could log the dropped data at this point, that may be a better solution.

tests/test_utils.py
```diff
@@ -184,3 +184,18 @@ def test_sanitize_list_handling(self):
...
         # no assert as we are just expecting this not to throw
         encoder._sanitize(data, ['password'], ['password'])
+
+    def test_json_encode_invalid_keys(self):
+        """
+        Test that _sanitize can accept some invalid json where a function
+        name or some other bad data is passed as a key in the payload dictionary.
+        """
+        def foo():
+            return "123"
```
Cawllec: We'll need to expand the number of test cases for other potentially invalid keys.

brockhaywood: Yea, I didn't have any immediate good ideas on other test cases where a key would be invalid but wouldn't be handled by the json encoder. For example, ints and None are handled internally and dont cause exceptions.

brockhaywood:
@Cawllec updated based on your feedback and tests are passing

brockhaywood:
@Cawllec can you re-review?

brockhaywood:
@Cawllec cool. updated the message and rebased against master

PR #166, opened 2018-11-30
Separate middleware stacks

Cawllec:
## Goal

Separate internal and external middleware stacks to ensure internal middleware is always run before external middleware.

## Tests

The individual changes have tests, the only things potentially needing more tests added are the integrations (Django/Flask/WSGI) that add middleware to the internal stack.  While the tests pass with checks that would fail if the middleware was in the wrong space, the don't specifically check for it.

### DIFF
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -66,7 +66,7 @@ def configure():
     django_bugsnag_settings = getattr(settings, 'BUGSNAG', {})
     bugsnag.configure(**django_bugsnag_settings)
 
-    bugsnag.legacy.configuration.internal_middleware.before_notify(add_django_request_to_notification)
+    bugsnag.configure().internal_middleware.before_notify(add_django_request_to_notification)
 
 
 def __track_session(sender, **extra):
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -66,7 +66,7 @@ def configure():
     django_bugsnag_settings = getattr(settings, 'BUGSNAG', {})
     bugsnag.configure(**django_bugsnag_settings)
 
-    bugsnag.before_notify(add_django_request_to_notification)
+    bugsnag.legacy.configuration.internal_middleware.before_notify(add_django_request_to_notification)
 
 
 def __track_session(sender, **extra):
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -66,7 +66,8 @@ def configure():
     django_bugsnag_settings = getattr(settings, 'BUGSNAG', {})
     bugsnag.configure(**django_bugsnag_settings)
 
-    bugsnag.configure().internal_middleware.before_notify(add_django_request_to_notification)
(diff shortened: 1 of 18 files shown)


bugsnag/middleware.py
```diff
@@ -104,6 +104,38 @@ def append(self, middleware):
...
+        else:
+            self.append(middleware)
+
+    def _find_class_index(self, target_class):
+        """
+        Locates the index of the target class in the stack.
+        Returns -1 if not present.
+        """
+        for index, item in enumerate(self.stack):
+            if item.__name__ == target_class:
```
tobyhs: It might be a better interface to have `insert_before` and `insert_after` take in actual class references/objects instead of strings since different classes can have the same `__name__` property.

bugsnag/middleware.py
```diff
@@ -104,6 +104,38 @@ def append(self, middleware):
         """
         self.stack.append(middleware)
 
+    def insert_before(self, middleware, target_class):
```
tobyhs: It might be more conventional to swap the order of the parameters so that the target comes before the middleware to add. This closer matches how bugsnag-ruby's middleware stack, Rails middleware stack, and Python's `list.insert` method work.

tests/test_notify.py
```diff
@@ -567,6 +567,116 @@ def callback(report):
...
+            notification.meta_data['test']['array'].append(1)
+
+        def second_callback(notification):
+            notification.meta_data['test']['array'].append(2)
+
+        # Simulate an internal middleware
+        bugsnag.legacy.configuration.internal_middleware.before_notify(
+                first_callback)
+        # Add a regular callback function
+        bugsnag.before_notify(second_callback)
```
tobyhs: Should you swap these so `bugsnag.before_notify` comes before `bugsnag.configuration.internal_middleware.before_notify` to ensure that internal middleware is still called first?

bugsnag/django/__init__.py
```diff
@@ -66,7 +66,7 @@ def configure():
     django_bugsnag_settings = getattr(settings, 'BUGSNAG', {})
     bugsnag.configure(**django_bugsnag_settings)
 
-    bugsnag.before_notify(add_django_request_to_notification)
+    bugsnag.legacy.configuration.internal_middleware.before_notify(add_django_request_to_notification)
```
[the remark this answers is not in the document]
Cawllec: This was due to an issue with the way the module is set up, the `configuration` in `legacy.py` isn't the same as what's returned from `bugsnag.configuration`, even though it looks like it should be.  We're discussing what to do with this, but it looks like we're favouring having `bugsnag.configure` return the configuration object, allowing access for middleware.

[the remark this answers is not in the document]
Cawllec: That was the middleware call will be `bugsnag.configure.internal_middleware.append`, which reads reasonably to me.

Cawllec:
I've updated based on the feedback from @tobyhs, so this could probably do with a re-review.

Cawllec:
Force push above was due to commit which should be in a separate branch.

PR #176, opened 2019-02-25
fix(flask): Guard against JSON parsing failing

kattrali:
If get_json fails, an exception can be thrown and cancel the callback
entirely.

## Changeset

* Silenced JSON parsing errors in Flask integration to ensure request data is appended successfully

## Tests

* Tested manually by constructing requests with errors and broken data with `Content-Type: application/json`
* Added an automated test in a similar setup

## Review

What this changeset needs most is thoroughness - is it possible to create the same situation in another way? I tried a few other options like manipulating the `request` object though all seem to be guarded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+## TBD
+
+### Fixes
+
+* Ensure `request` tab is attached to Flask requests when JSON data is malformed
+  or otherwise cannot be read at crash time
+  [#176](https://github.com/bugsnag/bugsnag-python/pull/176)
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+## TBD
+
+### Fixes
+
+* Ensure `request` tab is attached to Flask requests when JSON data is malformed
+  or otherwise cannot be read at crash time
+  [#175](https://github.com/bugsnag/bugsnag-python/pull/175)
```
tobyhs: I'm assuming this should be 176

PR #180, opened 2019-06-12
DO NOT MERGE - cut-down tests to show random FilterDict test failure

tomlongridge:
## Goal

In trying to add runtime information to the device payload I've added a nested dict to the structure and this seems to be failing unit tests sporadically. I have reduced the payload in a unit test down to the following:

```
        {
            "level1-key1": {
                "level2-key1": FilterDict({
                    "level3-key1": {
                        'level4-key1': 'level4-value1'}
                }),
                "level2-key2": FilterDict({
                    "level3-key2": "level3-value1",
                    "level3-key3": {
                        'level4-key2': 'level4-value2'}
                }),
            }
        }
```

It's the addition of the last dict that I'm trying to achieve and yet it doesn't occur if I remove the other FilterDict or the simple string pair preceding it.

I've added the `test_to_death_tx.sh` script that can be run to continuously run the tests - they fail after <10 usually. It only seems to fail through tox on python3.5 and before - running nosetest doesn't seem to make it fail.

I've added print lines to the utils function to show what's happening and on a failed test you can see two dict's added to the `ignored` list with the same `id` despite being different objects.

It is always a clash between `level4-key1` and `level4-key2`:

```
***** Added: 4552960392 -> {'level4-key2': 'level4-value2'}
```

```
!!!!! Recursive: 4552960392 -> {'level4-key1': 'level4-value1'}
```

I wondered whether it was to do with the order of dict iteration being undefined but it doesn't seem to fix it with a sorted list either.


tomlongridge:
Fixed in #181

PR #181, opened 2019-06-12
fix(utils): Make nested dict uniqueness checks work

kattrali:
## Goal

Fix potentially over-aggressive recursion checking where references are discarded between id() checks.

## Design

References to nested dicts/field are recycled when going out of scope, which is immediately since the references are not retained for the lifetime of the sanitization pass. This change keeps references to seen nested objects in addition to ignored objects to ensure id() returns unique references.

Related reports on the Python issue tracker which explain the issue:
* https://bugs.python.org/issue36156
* https://bugs.python.org/issue33685

```
>>> help(id)

Help on built-in function id in module __builtin__:

id(...)
    id(object) -> integer

    Return the identity of an object.  This is guaranteed to be unique
    among simultaneously existing objects.  (Hint: it's the object's 
    memory address.)
```

## Changeset

* Update `_sanitize` and `_sanitize_dict` encoder methods to keep seen references

## Tests

* Added a new test case which checks for recursion in a large enough dict
* Tested against `tox` config and underlying `nosetests` directly

## Linked issues

Related to #179 
Closes #180

## Review

### DIFF
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -66,7 +66,8 @@ class SanitizingJSONEncoder(JSONEncoder):
                 if is_string and any(f in key.lower() for f in self.filters):
                     clean_dict[key] = self.filtered_value
                 else:
-                    clean_dict[key] = self.filter_string_values(value, ignored, seen)
+                    clean_dict[key] = self.filter_string_values(
+                        value, ignored, seen)
 
             return clean_dict
 
@@ -96,7 +97,7 @@ class SanitizingJSONEncoder(JSONEncoder):
             ignored = set()
 
         # Keep track of nested objects to avoid having references garbage
-        # collected (which would cause id reuse and false positive recursion
+        # collected (which would cause id reuse and false positive recursion)
         if seen is None:
             seen = []
 
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -38,12 +38,17 @@ class SanitizingJSONEncoder(JSONEncoder):
         else:
             return payload
 
-    def filter_string_values(self, obj, ignored=None):
+    def filter_string_values(self, obj, ignored=None, seen=None):
         """
         Remove any value from the dictionary which match the key filters
         """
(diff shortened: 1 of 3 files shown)


bugsnag/utils.py
```diff
@@ -123,7 +128,7 @@ def _sanitize_dict_key_value(self, clean_dict, key, clean_value):
                     'Could not add sanitize key for dictionary, '
                     'dropping value.')
 
-    def _sanitize_dict(self, obj, trim_strings, ignored):
+    def _sanitize_dict(self, obj, trim_strings, ignored, seen):
```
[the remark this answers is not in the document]
kattrali: yeah its intentional, since filtering does its own "ignored" checking as its available at the top level. Though it should also do its own "seen" checking!

bugsnag/utils.py
```diff
@@ -81,13 +81,15 @@ def default(self, obj):
...
 
-    def _sanitize(self, obj, trim_strings, ignored=None):
+    def _sanitize(self, obj, trim_strings, ignored=None, seen=None):
         """
         Replace recursive values and trim strings longer than
         MAX_STRING_LENGTH
         """
         if not ignored:
             ignored = set()
+        if not seen:
```
tobyhs: It seems preferable to compare `seen` to `None` directly, as `not []` is true, so it would create another empty array unnecessarily. However, I can't see how `_sanitize` will be called with an empty `seen` with the current code, so it probably doesn't matter too much.

bugsnag/utils.py
```diff
@@ -81,13 +81,15 @@ def default(self, obj):
...
+            seen = []
```
tobyhs: Consider adding a comment to explain why this `seen` parameter is needed, as it probably isn't obvious to someone reading the code without looking at the pull request description or commit message.

kattrali:
Thanks for the review, @tobyhs, those sound like nice additions. Updated the PR to match.

PR #182, opened 2019-06-18
feat: add runtime version to framework integrations

tomlongridge:
## Goal

Adds version information for each integration with bottle being a conditional import
within wsgi. Also added a bottle example project.

### Added

- Bottle example project

### Changed

- celery/__init__.py - to add celery.__version__
- django/__init__.py - to add django.__version__
- flask/__init__.py - to add flask.__version__
- tornado/__init__.py - to add tornado.version
- wsgi/middleware.py - to add bottle.__version__ (if available)

## Tests

Re-ran example projects to ensure deviceData was populated as expected.

## Review

  - [x] Final review

### DIFF
--- a/bugsnag/celery/__init__.py
+++ b/bugsnag/celery/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import division, print_function, absolute_import
 
+import celery
 from celery.signals import task_failure
 import bugsnag
 
@@ -28,4 +29,5 @@ def connect_failure_handler():
     Connect the bugsnag failure_handler to the Celery
     task_failure signal
     """
+    bugsnag.configure().runtime_versions['celery'] = celery.__version__
     task_failure.connect(failure_handler, weak=False)
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import division, print_function, absolute_import
 
 import six
+import django
 from django.conf import settings
 from django.core.signals import request_started
 
@@ -68,6 +69,8 @@ def configure():
 
     middleware = bugsnag.configure().internal_middleware
     middleware.before_notify(add_django_request_to_notification)
+
+    bugsnag.configure().runtime_versions['django'] = django.__version__
 
 
 def __track_session(sender, **extra):
--- a/bugsnag/tornado/__init__.py
+++ b/bugsnag/tornado/__init__.py
@@ -1,3 +1,4 @@
+import tornado
 from tornado.web import RequestHandler
 from tornado.web import HTTPError
 
@@ -38,6 +39,7 @@ class BugsnagRequestHandler(RequestHandler):
         RequestHandler._handle_request_exception(self, exc)
 
     def prepare(self):
+        bugsnag.configure().runtime_versions['tornado'] = tornado.version
         if bugsnag.configuration.auto_capture_sessions:
             bugsnag.start_session()
 
--- a/bugsnag/wsgi/middleware.py
+++ b/bugsnag/wsgi/middleware.py
@@ -1,13 +1,17 @@
 from webob import Request
+import sys
 
 import bugsnag
 from bugsnag.wsgi import request_path
 
-try:
-    import bottle
+bottle_present = False
+if 'bottle' in sys.modules:
     bottle_present = True
-except ImportError:
-    bottle_present = False
+    try:
+        import bottle
+    except ImportError:
+        bottle_present = False
+
 
 def add_wsgi_request_data_to_notification(notification):
     if not hasattr(notification.request_config, "wsgi_environ"):
@@ -42,9 +46,6 @@ class WrappedWSGIApp(object):
         self.environ = environ
 
         bugsnag.configure_request(wsgi_environ=self.environ)
-        if bottle_present:
-            bugsnag.configure().runtime_versions['bottle'] = bottle.__version__
-
         try:
             if bugsnag.configuration.auto_capture_sessions:
                 bugsnag.start_session()
@@ -93,6 +94,8 @@ class BugsnagMiddleware(object):
 
     def __init__(self, application):
         middleware = bugsnag.configure().internal_middleware
+        if bottle_present:
+            bugsnag.configure().runtime_versions['bottle'] = bottle.__version__
         middleware.before_notify(add_wsgi_request_data_to_notification)
         self.application = application
 
--- a/bugsnag/wsgi/middleware.py
+++ b/bugsnag/wsgi/middleware.py
@@ -3,6 +3,11 @@ from webob import Request
 import bugsnag
 from bugsnag.wsgi import request_path
 
+try:
+    import bottle
+    bottle_present = True
+except ImportError:
+    bottle_present = False
 
 def add_wsgi_request_data_to_notification(notification):
     if not hasattr(notification.request_config, "wsgi_environ"):
@@ -37,6 +42,8 @@ class WrappedWSGIApp(object):
         self.environ = environ
 
         bugsnag.configure_request(wsgi_environ=self.environ)
+        if bottle_present:
+            bugsnag.configure().runtime_versions['bottle'] = bottle.__version__
 
         try:
             if bugsnag.configuration.auto_capture_sessions:
--- a/example/bottle/README.md
+++ b/example/bottle/README.md
@@ -1,6 +1,6 @@
-# Bugsnag Flask demo
+# Bugsnag Bottle demo
 
-This Flask application demonstrates how to use Bugsnag with the Flask web framework for Python.
+This Bottle application demonstrates how to use Bugsnag with the Bottle web framework for Python.
 
 ## Setup
 
@@ -11,7 +11,7 @@ To get set up, follow the instructions below. Don't forget to replace the placeh
 1. Clone the repo and `cd` into this directory:
     ```sh
     git clone https://github.com/bugsnag/bugsnag-python.git
-    cd bugsnag-python/example/flask
+    cd bugsnag-python/example/bottle
     ```
 
 1. Install dependencies
@@ -27,7 +27,7 @@ file and configure your API key.
     python server.py
     ```
 
-1. View the example page which will (most likely) be served at: http://localhost:3000
+1. View the example page which will be served at: http://localhost:8080
 
 For more information, see our documentation:
-https://docs.bugsnag.com/platforms/python/flask/
+https://docs.bugsnag.com/platforms/python/bottle/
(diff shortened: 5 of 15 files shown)


example/bottle/server.py
```diff
@@ -0,0 +1,32 @@
...
+    """Deliberately raises an unhandled error and crash the app.
+    """
+    raise(Exception('SomethingBad'))
+
+@route('/handled')
+def handled():
+    try:
+        x = 1/0
+    except ZeroDivisionError:
+        bugsnag.notify(ZeroDivisionError('Bottle demo: To infinity... and beyond!'))
```
tobyhs: Something seems off here. You catch the `ZeroDivisionError` but don't use it.

It seems like this code do one of the following:
1) have `except ZeroDivisionError as e:` on the previous line, then call `bugsnag.notify(e)`
2) Avoid the `try`+`except` and just call `bugsnag.notify(ZeroDivisionError(...))`

bugsnag/wsgi/middleware.py
```diff
@@ -37,6 +42,8 @@ def __init__(self, application, environ, start_response):
         self.environ = environ
 
         bugsnag.configure_request(wsgi_environ=self.environ)
+        if bottle_present:
+            bugsnag.configure().runtime_versions['bottle'] = bottle.__version__
```
tobyhs: Should this be moved to `BugsnagMiddleware.__init__` instead?

It seems like setting the bottle version only needs to happen once, but I think `WrappedWSGIApp.__init__` is called for every request.

bugsnag/django/__init__.py
```diff
@@ -69,6 +70,8 @@ def configure():
     middleware = bugsnag.configure().internal_middleware
     middleware.before_notify(add_django_request_to_notification)
 
+    bugsnag.configure().runtime_versions['django'] = django.__version__
```
tobyhs: Should there be something in tests/integrations/test_django_1.py that checks for the presence of this entry? (similar question applies to the Flask integration too)

example/bottle/server.py
tomlongridge: I have just lifted the functions from the `flask` example, but I take your point. I've amended both to match the `plain` example.

bugsnag/wsgi/middleware.py
tomlongridge: Good spot - thanks, I hadn't noticed that.

bugsnag/celery/__init__.py
```diff
@@ -28,4 +29,5 @@ def connect_failure_handler():
     Connect the bugsnag failure_handler to the Celery
     task_failure signal
     """
+    bugsnag.configure().runtime_versions['celery'] = celery.__version__
```
kylef: I think this change is missing test coverage.

bugsnag/tornado/__init__.py
```diff
@@ -38,6 +39,7 @@ def _handle_request_exception(self, exc):
         RequestHandler._handle_request_exception(self, exc)
 
     def prepare(self):
+        bugsnag.configure().runtime_versions['tornado'] = tornado.version
```
kylef: I think this change is missing test coverage.

example/django/requirements.txt
```diff
@@ -1,2 +1,2 @@
-Django>=2.0.1
+Django==2.0.1
```
kylef: Why would we want to pin this example to an outdated version patch of Django 2? This particular version of Django has known vulnerabilities registered as CVEs.

example/flask/server.py
```diff
@@ -100,16 +100,15 @@ def crashcallback():
...
 
 @app.route('/handled')
-def notify():
+def handle_zero_div():
     """Deliberately triggers a handled exception, and reports it to Bugsnag.
     """
     try:
         x = 1/0
-    except ZeroDivisionError:
-        bugsnag.notify(ZeroDivisionError('Flask demo: To infinity... and beyond!'))
```
kylef: This change makes this particular demo inconsistent with the rest of the examples (Django 2, Django 1, celery). What's the justification for that (and how is that related to the runtime version)?

tests/integrations/test_flask.py
```diff
@@ -211,3 +212,21 @@ def test():
...
+        handle_exceptions(app)
+        app.test_client().get('/hello')
+
+        self.assertEqual(len(self.server.received), 1)
+
+        payload = self.server.received[0]['json_body']
+        device_data = payload['events'][0]['device']
+
+        self.assertEqual(device_data['runtimeVersions']['flask'],
+                         flask.__version__)
```
kylef: Perhaps it would be a good idea to test a bit more details about the version set here, the test lacks protection against changes in flask regarding the `__version__` and gives a false sense of security.

For example, if in the next major version of Flask the API of `__version__` changes to be an array of numbers or an object (example: `__version__ = {'major': 18, 'minor': 0, 'patch': 0}`) this test would continue to pass after upgrading flask. That means anyone in the future going to ensure that the integration works with a new version of Flask (the same applies to the other integrations) may wrongly expect that the code is compatible with the newer version. Which isn't correct, it would unexpectedly change the format of the version sent to the server.

It may be a good idea to also test that `device_data['runtimeVersions']['flask']` matches a regex like `\d+\.\d+\.\d+` which ensures it is a string of the expected format?

bugsnag/wsgi/middleware.py
```diff
@@ -1,8 +1,17 @@
 from webob import Request
+import sys
 
 import bugsnag
 from bugsnag.wsgi import request_path
 
+bottle_present = False
+if 'bottle' in sys.modules:
```
kylef: Perhaps a little more idomatic to use:

```python
try:
    import bottle
except ImportError:
    bottle = None

if bottle:
    bugsnag.configure().runtime_versions['bottle'] = bottle.__version__
```

bugsnag/wsgi/middleware.py
```diff
@@ -1,8 +1,17 @@
...
+    bottle_present = True
+    try:
+        import bottle
```
kylef: I wonder if this could give false positives, just because `bottle` is importable does not mean the code base is using it.

Bottle could be installed as a global dependency on a server (perhaps on a shared web hosted, such as `apt-get install python-bottle`), there may be other Python applications using bugsnag-python running on that machine which would incorrectly report they are using bottle of a given version. I am wondering more about shared hosting providers.

example/bottle/README.md
```diff
@@ -0,0 +1,33 @@
...
+This Bottle application demonstrates how to use Bugsnag with the Bottle web framework for Python.
+
+## Setup
+
+Try this out with [your own Bugsnag account](https://app.bugsnag.com/user/new)! You'll be able to see how the errors are reported in the dashboard, how breadcrumbs are left, how errors are grouped and how they relate to the original source.
+
+To get set up, follow the instructions below. Don't forget to replace the placeholder API token with your own!
+
+1. Clone the repo and `cd` into this directory:
+    ```sh
```
kylef:
```suggestion
    ```shell
```

You're using `shell` everywhere else in this file.

example/bottle/README.md
```diff
@@ -0,0 +1,33 @@
+# Bugsnag Bottle demo
```
kylef: This new integration example does not seem to have any tests.

example/bottle/index.html
```diff
@@ -0,0 +1,31 @@
...
+    <div id="python">
+        <a href="/crash" style="text-decoration:none;">
+          <button type="button">UNHANDLED ERROR</button>
+        </a>
+        <br>
+        <a href="/handled" style="text-decoration:none;">
+          <button type="button">HANDLED ERROR</button>
+        </a>
+    </div>
+<!-- review settings.py & views.py to see how bugsnag integrates with your app. -->
```
kylef: Should this be a comment? Perhaps it would be a good idea to make this an actual paragraph so readers of the rendered HTML page can view the insight.

example/bottle/server.py
```diff
@@ -0,0 +1,34 @@
...
+
+@route('/')
+def index():
+    return static_file('index.html', root='.')
+
+@route('/crash')
+def crash():
+    """Deliberately raises an unhandled error and crash the app.
+    """
+    raise(Exception('SomethingBad'))
```
kylef: raise is not a function

```suggestion
    raise Exception('SomethingBad')
```

also flake8 is failing on this file:

```
$ flake8 example/bottle/server.py
example/bottle/server.py:10:1: E302 expected 2 blank lines, found 1
example/bottle/server.py:14:1: E302 expected 2 blank lines, found 1
example/bottle/server.py:20:1: E302 expected 2 blank lines, found 1
example/bottle/server.py:25:9: F841 local variable 'x' is assigned to but never used
example/bottle/server.py:25:14: E226 missing whitespace around arithmetic operator
example/bottle/server.py:29:80: E501 line too long (93 > 79 characters)
example/bottle/server.py:31:1: E305 expected 2 blank lines after class or function definition, found 1
```

tests/integrations/test_flask.py
tomlongridge: Good point - I'll amend.

bugsnag/wsgi/middleware.py
tobyhs: I think that was what line 8 (`if 'bottle' in sys.modules`) is supposed to partially account for (if the user imported bottle, then they are likely using it).

example/flask/server.py
tobyhs: It wasn't related to runtime version directly; it was related to a previous comment in this pull request (https://github.com/bugsnag/bugsnag-python/pull/182#discussion_r294985483) for the new bottle example. I would have preferred that similar changes to different example apps be separate changes or pull requests, but from my previous jobs/companies, fighting every battle that I wanted was too mentally exhausting and lead to me being heavily criticized as "too picky" (and other similar things), so I had to pick my battles.

bugsnag/wsgi/middleware.py
tomlongridge: Indeed, I believe this will only trigger if bottle is imported elsewhere.

example/bottle/README.md
tomlongridge: This isn't a new integration as such - it uses wsgi - I just added an extra example whilst I was testing the runtime versioning.

bugsnag/tornado/__init__.py
tomlongridge: It is, so the change was manually tested using the example. I've raised an internal ticket to consider adding unit tests to this integration.

bugsnag/celery/__init__.py
tomlongridge: It is, so the change was manually tested using the example. I've raised an internal ticket to consider adding unit tests to this integration.

example/flask/server.py
tomlongridge: I agree with our comments @tobyhs on #182 - the examples are already a little inconsistent and break style rules, so I've raised a ticket to sort this out separately.

PR #183, opened 2019-06-21
Runtime version frameworks additional tweaks

tomlongridge:
## Goal

Following #182, a few further tweaks to runtime versions following review comments after merge.

### DIFF
--- a/example/bottle/server.py
+++ b/example/bottle/server.py
@@ -4,18 +4,21 @@ import bugsnag
 from bugsnag.wsgi.middleware import BugsnagMiddleware
 
 bugsnag.configure(
-    api_key='698ffb05632f0ffcad3ebd16383507fd',
+    api_key='some-api-key',
 )
+
 
 @route('/')
 def index():
     return static_file('index.html', root='.')
 
+
 @route('/crash')
 def crash():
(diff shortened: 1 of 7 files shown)


example/bottle/server.py
```diff
@@ -26,7 +29,9 @@ def handle_zero_div():
     except Exception as e:
```
tobyhs: (GitHub won't let me comment one line higher)

flake8 on this file is complaining about `x` being an unused local variable, and this seems to be the only lint issue left on this file. Perhaps this should be fixed and then `example/bottle` can be added to scripts/lint.sh.

tomlongridge: I've raised an internal ticket to address this - suggest not doing it here as it affects the other examples too.

example/bottle/server.py
```diff
@@ -4,18 +4,21 @@
 from bugsnag.wsgi.middleware import BugsnagMiddleware
 
 bugsnag.configure(
-    api_key='698ffb05632f0ffcad3ebd16383507fd',
+    api_key='some-api-key',
```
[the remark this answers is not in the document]
tomlongridge: Added this to examples ticket for future work.

PR #187, opened 2020-05-04
Test on Python 3.7 & 3.8 and speed up Travis

imjoehaines:
## Goal

Due to changes in Travis' build images and Python package updates, the Python 2.6 and 3.3 jobs on Travis don't work in master

This PR fixes them by using `dist: trusty` so they are installable again and pinning dependencies on 3.3

I've also collapsed all of the jobs into one job per Python version, which speeds up the test suite from taking [~35 minues](https://travis-ci.com/github/bugsnag/bugsnag-python/builds/163724474) to taking [~13 minutes](https://travis-ci.com/github/bugsnag/bugsnag-python/builds/163732556). This makes the Travis file a bit gross, but AFAICT that's the only downside — tox still runs each in a separate environment and coveralls reports the same coverage

### Changed

- Travis is now running one build per Python version
- Python 3.7 and 3.8 are running on Travis
- tox no longer lists environments which aren't valid

## Review

  - [x] Final review

### DIFF
--- a/tests/test_delivery.py
+++ b/tests/test_delivery.py
@@ -1,10 +1,5 @@
-import sys
-
-from nose.plugins.skip import SkipTest
-
 from bugsnag import Configuration
-from bugsnag.delivery import (UrllibDelivery, RequestsDelivery,
-                              create_default_delivery)
+from bugsnag.delivery import (UrllibDelivery, create_default_delivery)
 
 from tests.utils import IntegrationTest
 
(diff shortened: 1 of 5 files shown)


tests/test_delivery.py
```diff
@@ -36,40 +37,6 @@ def test_urllib_delivery(self):
...
-            raise SkipTest("Requests is not installed")
-
-        self.config.configure(endpoint=self.server.url)
-        del self.config.use_ssl
-        RequestsDelivery().deliver(self.config, '{"legit": 4}')
-
-        self.assertSentReportCount(1)
-        request = self.server.received[0]
-        self.assertEqual(request['json_body'], {'legit': 4})
-
```
imjoehaines, line 72: These deleted tests now live in `tests/requests/test_requests_delivery.py`

tests/test_notify.py
```diff
@@ -168,15 +168,16 @@ def test_notify_user_filter(self):
         self.assertEqual('Man', event['user']['lastname'])
 
     def test_notify_payload_matching_filter(self):
-        bugsnag.configure(params_filters=['number'])
-        bugsnag.notify(ScaryException('unexpected failover'),
-                       apple='four', number=76)
+        bugsnag.configure(params_filters=['bar'])
+        bugsnag.notify(ScaryException('unexpected failover'), foo=4, bar=76)
```
imjoehaines, line 172: This change is to pull all of the `bugsnag.notify` call onto one line, which gets around a small difference in how Python 3.8 reports line numbers in stack traces — see b4f3095664456572af1b83da691bded0f13222fb and https://bugs.python.org/issue12458 for more information

I think this is a bit nicer than doing a dynamic version check to see if the line number should be 172 or 173

tox.ini
```diff
@@ -1,10 +1,13 @@
...
+    py{26,27,33,34,35,36,37,38}-requests,
     py{26,27}-flask,
-    py{26,27,33,34,35,36}-wsgi,
-    py{27,36}-django{18,19,110,111,20,21}-migrate1-django1
-    py36-lint,
+    py{26,27,33,34,35,36,37,38}-wsgi,
+    py{35,36,37}-django{18,19,110,111,20,21}-migrate1-django1,
+    py27-django{18,19,110,111}-migrate1-django1,
+    py38-django{18,111,20,21}-migrate1-django1,
+    py38-lint
```
imjoehaines, line 10: I've split out these to remove combinations that aren't valid, e.g. `py27-django20-migrate1-django1` doesn't work so it's not longer here

PR #188, opened 2020-05-04
Prevent exceptions when given badly encoded URLs

### LINKED ISSUE #185: WSGI middleware breaks when handling badly encoded URLs
### Description

In public-facing applications, we cannot control what kinds of URLs will be requested by clients -- which means we may have to deal with wrong or broken encoding.

### Issue

`bugsnag.wsgi.middleware.add_wsgi_request_data_to_notification` uses `webob.Request.path_url`, which raises an exception if the URL path is encoded improperly. It would not break if it instead e.g. used the already existing helper function `bugsnag.wsgi.request_path`.

### Environment

Library versions:

- python version: both 2.7 and 3.7
- bugsnag-python version: 3.6.0

### Example code snippet

```python
import unittest
import unittest.mock
import urllib.parse

class PathEncodingTest(unittest.TestCase):

    environ = {
        # https://github.com/python/cpython/blob/master/Lib/wsgiref/simple_server.py#L85
        'PATH_INFO': urllib.parse.unquote('/%83', 'latin-1'),
        'SCRIPT_NAME': '',
        'SERVER_NAME': 'localhost',
        'SERVER_PORT': '80',
        'wsgi.url_scheme': 'http',
    }

    def test_wrongly_encoded_url_should_not_raise(self):
        import bugsnag.wsgi.middleware
        notification = unittest.mock.Mock()
        notification.request_config.wsgi_environ = self.environ
        # This currently raises UnicodeDecodeError
        bugsnag.wsgi.middleware.add_wsgi_request_data_to_notification(
            notification)
        # but I'd expect something like this:
        self.assertEqual(
            'http://localhost/%83',
            notification.add_tab.call_args_list[0][0][1]['url'])

    def test_using_the_already_existing_helper_function_works_somewhat(self):
        import bugsnag.wsgi
        import webob
        r = webob.Request(self.environ)
        self.assertEqual(
            # Not sure how correct this is, but at least it doesn't break
            'http://localhost/%C2%83',
            r.application_url + bugsnag.wsgi.request_path(self.environ))

    def test_werkzeug_instead_of_webob_also_handles_it_without_error(self):
        import werkzeug
        r = werkzeug.Request(self.environ)
        self.assertEqual('http://localhost/%83', r.base_url)

if __name__ == '__main__':
    unittest.main()
```

<details><summary>Error messages:</summary>

```
..E
======================================================================
ERROR: test_wrongly_encoded_url_should_not_raise (__main__.EncodingTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "../foo.py", line 21, in test_wrongly_encoded_url_should_not_raise
    notification)
  File "/Users/schnerring/Downloads/venv/lib/python3.7/site-packages/bugsnag/wsgi/middleware.py", line 28, in add_wsgi_request_data_to_notification
    "url": request.path_url,
  File "/Users/schnerring/Downloads/venv/lib/python3.7/site-packages/webob/request.py", line 467, in path_url
    bpath_info = bytes_(self.path_info, self.url_encoding)
  File "/Users/schnerring/Downloads/venv/lib/python3.7/site-packages/webob/descriptors.py", line 70, in fget
    return req.encget(key, encattr=encattr)
  File "/Users/schnerring/Downloads/venv/lib/python3.7/site-packages/webob/request.py", line 165, in encget
    return bytes_(val, 'latin-1').decode(encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x83 in position 1: invalid start byte

----------------------------------------------------------------------
Ran 3 tests in 0.099s

FAILED (errors=1)
```
</details>

imjoehaines:
## Goal

Webob will raise exceptions if given a URL it cannot decode. This is probably a good thing for applications, but we should be more resilient

We already have a helper for getting the path of a URL which handles bad encoding using `urllib`

### Linked issues

Fixes #185

## Review

- [x] A changelog entry was added for the goal of this pull request
  - [x] Final review

### DIFF
--- a/bugsnag/wsgi/middleware.py
+++ b/bugsnag/wsgi/middleware.py
@@ -21,11 +21,12 @@ def add_wsgi_request_data_to_notification(notification):
 
     environ = notification.request_config.wsgi_environ
     request = Request(environ)
+    path = request_path(environ)
 
-    notification.context = "%s %s" % (request.method, request_path(environ))
+    notification.context = "%s %s" % (request.method, path)
     notification.set_user(id=request.remote_addr)
     notification.add_tab("request", {
-        "url": request.path_url,
+        "url": request.application_url + path,
         "headers": dict(request.headers),
         "params": dict(request.params),
     })
--- a/bugsnag/wsgi/middleware.py
+++ b/bugsnag/wsgi/middleware.py
@@ -26,7 +26,7 @@ def add_wsgi_request_data_to_notification(notification):
     notification.context = "%s %s" % (request.method, path)
     notification.set_user(id=request.remote_addr)
     notification.add_tab("request", {
-        "url": request.application_url + path,
+        "url": "%s%s" % (request.application_url, path),
         "headers": dict(request.headers),
         "params": dict(request.params),
     })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Fixes
+
+* WSGI middleware will no longer raise when given a badly encoded URL
+  [#188](https://github.com/bugsnag/bugsnag-python/pull/188)
+
 ## 3.6.0 (2019-06-25)
 
 ### Enhancements
--- a/tests/test_path_encoding.py
+++ b/tests/test_path_encoding.py
@@ -1,6 +1,10 @@
+# coding=utf-8
+
 import unittest
-import unittest.mock
-import urllib.parse
+from six.moves import urllib
+
+from bugsnag.notification import Notification
+from bugsnag.configuration import (Configuration, RequestConfiguration)
 
 
 class PathEncodingTest(unittest.TestCase):
@@ -17,8 +21,14 @@ class PathEncodingTest(unittest.TestCase):
         environ = self.environ.copy()
         environ['PATH_INFO'] = '/hello/world'
 
-        notification = unittest.mock.Mock()
-        notification.request_config.wsgi_environ = environ
+        bugsnag.configure_request(wsgi_environ=environ)
+
+        config = Configuration()
+        notification = Notification(
+            Exception("oops"),
+            config,
+            RequestConfiguration.get_instance()
+        )
 
         bugsnag.wsgi.middleware.add_wsgi_request_data_to_notification(
             notification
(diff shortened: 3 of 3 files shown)


bugsnag/wsgi/middleware.py
```diff
@@ -21,11 +21,12 @@ def add_wsgi_request_data_to_notification(notification):
...
     environ = notification.request_config.wsgi_environ
     request = Request(environ)
+    path = request_path(environ)
 
-    notification.context = "%s %s" % (request.method, request_path(environ))
+    notification.context = "%s %s" % (request.method, path)
     notification.set_user(id=request.remote_addr)
     notification.add_tab("request", {
-        "url": request.path_url,
+        "url": request.application_url + path,
```
tomlongridge, line 29: Nit: perhaps use string format syntax to match above?

tomlongridge, line 29: Do we know that `application_url` is safe to use?

imjoehaines: I'll do some reading to make sure but I assume it will be safe because if it's badly encoded I don't think you would be able reach the webserver in the first place, though I'm not sure if this is possible to pass from the client

imjoehaines: I'm not sure about this to be honest — the `application_url` will be built using the `Host` header (if present), which is what's recommended in [the WSGI PEP](https://www.python.org/dev/peps/pep-3333/#url-reconstruction) but means that it's sent from the client so could have the same issues

Usually though the `Host` header would be set by the webserver talking to Python so the client version would be ignored, but that's not a guarantee so this could have the same problem

<details>
<summary>Github, for example, will try to redirect you if you send a bad `Host`</summary>

```
$ curl 'https://github.com' -H 'Host: google.com' -v
...
< HTTP/1.1 301 Moved Permanently
< Content-length: 0
< Location: https://github.com/
...
```

</details>

<details>
<summary>Python.org rejects the request completely</summary>

```
$ curl https://www.python.org -H 'Host: google.com'
<html>
<head>
<title>Fastly error: unknown domain google.com</title>
</head>
<body>
<p>Fastly error: unknown domain: google.com. Please check that this domain has been added to a service.</p>
</body>
</html>
```
</details>

Does it make sense to implement the [URL reconstruction algorithm](https://www.python.org/dev/peps/pep-3333/#url-reconstruction) ourselves, ignoring `HTTP_HOST`, if accessing `request.application_url` fails?

I don't think we can change the default of `request.application_url` because without the `Host` header, requests that are fronted by something like NGINX will be to `localhost` or an internal IP rather than the actual URL

tomlongridge: I think on balance what you have is appropriate as it's much less likely that the host headers will be badly encoded and different from the actual URL used to reach the server in the first place.

imjoehaines: 👍 7c96a25

PR #190, opened 2020-05-22
Run Flask tests on Python 3

imjoehaines:
Currently we're only running the Flask tests on Python 2, but they pass as-is on all our supported Python 3 versions so we can enable them everywhere

We also weren't running all of the test suites on Python 3.3. I managed to get this working locally (thanks to [this workaround](https://github.com/pyenv/pyenv/issues/1579#issuecomment-619594433)) and the same suites that we run on Python 3.4 pass, so I've enabled those too

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 envlist=
     py{26,27,33,34,35,36,37,38}-test,
     py{26,27,33,34,35,36,37,38}-requests,
-    py{26,27}-flask,
+    py{26,27,33,34,35,36,37,38}-flask,
     py{26,27,33,34,35,36,37,38}-wsgi,
     py{35,36,37}-django{18,19,110,111,20,21}-migrate1-django1,
     py27-django{18,19,110,111}-migrate1-django1,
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -2,7 +2,7 @@
 envlist=
     py{26,27,33,34,35,36,37,38}-test,
     py{26,27,33,34,35,36,37,38}-requests,
-    py{26,27}-flask,
+    py{26,27,33,34,35,36,37,38}-flask,
```
twometresteve, line 5: Mention of 33 here but not 3.3 in the yml file.  Is that significant?

imjoehaines: Travis didn't like running tox with Python 3.3 (e.g. https://travis-ci.com/github/bugsnag/bugsnag-python/jobs/338799201)

I figured it was nicer to define them here to show what could be run on 3.3 and they pass for me locally on 3.3 via tox

I think I gave up too easily getting it to run on Travis though, I'll spend a little bit more time trying to make it work more like the other versions

imjoehaines: Yeah, I definitely gave up too easily — it's now running with Tox on Travis :) Still a bit of a special case, but better than it was!

PR #192, opened 2020-07-13
feat: add request body for JSON POST requests

steve-nester-uk:
## Goal
The Bugsnag Django and Tornado notifiers currently only handles application/x-www-form-urlencoded POST data.  We want to handle application/json http body.

## Changeset
update to add_django_request_to_notification method to add request.body to the request meta data in case of application/JSON POST request.

## Tests
Tested manually using:
```
curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"json":"testdata","test":"xyz"}' \
  http://localhost:8000/app/
```
Added test_notify_post to test application/www-x-form-urlencoded POST
Added test_notify_json_post to test application/json POST
Added tests for tornado

## Review

- [ x] Commented on code changes inline explain the reasoning behind the approach
- [ x] Reviewed the test cases added for completeness and possible points for discussion
- [ x] Check the scope of the changeset - is everything in the diff required for the pull request?
  - [ x] Structural review of the classes, functions, and properties modified
  - [ x] Final review


bugsnag/django/__init__.py
```diff
@@ -45,14 +46,23 @@ def add_django_request_to_notification(notification):
...
         'method': request.method,
         'path': request.path,
         'encoding': request.encoding,
         'GET': dict(request.GET),
         'POST': dict(request.POST),
         'url': request.build_absolute_uri(),
-    })
+    }
+    if 'CONTENT_TYPE' in request.META:
+        if request.META['CONTENT_TYPE'] == 'application/json':
```
imjoehaines, line 58: This could be simplified a bit:

```suggestion
    if request.META.get('CONTENT_TYPE') == 'application/json':
```

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,23 @@ def add_django_request_to_notification(notification):
 
     if getattr(request, "session", None):
         notification.add_tab("session", dict(request.session))
-    notification.add_tab("request", {
+    req = {
```
imjoehaines, line 49: Is `request_tab` a better name for this?

tests/integrations/test_django_1.py
```diff
@@ -44,6 +44,38 @@ def test_notify(self):
...
+            'method': 'POST',
+            'url': 'http://testserver/notify/',
+            'path': '/notify/',
+            'POST': {'test': ['post']},
+            'encoding': None,
+            'GET': {}
+        })
+
+    def test_notify_json_post(self):
+        response = self.client.post('/notify/', {"test": "post"},
```
imjoehaines, line 63: I think this is what's causing the need for the string replace — this parameter isn't being converted to valid JSON because it's not a string:

```suggestion
        response = self.client.post('/notify/', '{"test": "post"}',
```

Django seems to just converts the dict to a string, which isn't valid JSON because it has single quotes around the keys/values:

```
>>> str({"test": "post"})
"{'test': 'post'}"
```

The string replace will break if the JSON has a `'` in it:

```
>>> import json
>>> body = """{"key": "val'ue"}"""
>>> body
'{"key": "val\'ue"}'
>>> json.loads(body)
{'key': "val'ue"}
>>> replaced = body.replace("'", '"')
>>> replaced
'{"key": "val"ue"}'
>>> json.loads(replaced)
Traceback (most recent call last):
...
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 14 (char 13)
```

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,23 @@ def add_django_request_to_notification(notification):
...
         'encoding': request.encoding,
         'GET': dict(request.GET),
         'POST': dict(request.POST),
         'url': request.build_absolute_uri(),
-    })
+    }
+    if 'CONTENT_TYPE' in request.META:
+        if request.META['CONTENT_TYPE'] == 'application/json':
+            req["content_type"] = request.META['CONTENT_TYPE']
+            if req["method"] == "POST":
```
imjoehaines, line 60: This block probably needs to be wrapped in a `try`, so we don't crash if the body doesn't decode or the JSON fails to parse

I think you can test this by sending some nonsense JSON and checking that POST is empty. Without a `try` here, the whole request tab will empty if this fails

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,23 @@ def add_django_request_to_notification(notification):
...
+                body = request.body.decode('utf-8')
+                uo = json.loads(body.replace("'", '"'))
+                req["POST"] = uo
```
imjoehaines, line 63: I'm not sure what `uo` stands for, but I think we can drop it anyway:

```suggestion
                req["POST"] = json.loads(body)
```

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,23 @@ def add_django_request_to_notification(notification):
...
         'path': request.path,
         'encoding': request.encoding,
         'GET': dict(request.GET),
         'POST': dict(request.POST),
         'url': request.build_absolute_uri(),
-    })
+    }
+    if 'CONTENT_TYPE' in request.META:
+        if request.META['CONTENT_TYPE'] == 'application/json':
+            req["content_type"] = request.META['CONTENT_TYPE']
```
imjoehaines, line 59: I think we can omit this as we don't usually track it in other notifiers

bugsnag/django/__init__.py
steve-nester-uk: Fixed in 4b59d6cf5ee86e11ede693f10421c1f95379d044

bugsnag/django/__init__.py
steve-nester-uk: Fixed in 4b59d6cf5ee86e11ede693f10421c1f95379d044

tests/integrations/test_django_1.py
steve-nester-uk: Fixed in 4b59d6cf5ee86e11ede693f10421c1f95379d044

bugsnag/django/__init__.py
steve-nester-uk: Fixed in 4b59d6cf5ee86e11ede693f10421c1f95379d044

bugsnag/django/__init__.py
steve-nester-uk: Fixed in 4b59d6cf5ee86e11ede693f10421c1f95379d044

bugsnag/django/__init__.py
steve-nester-uk: Fixed in 4b59d6cf5ee86e11ede693f10421c1f95379d044

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,23 @@ def add_django_request_to_notification(notification):
...
         'method': request.method,
         'path': request.path,
         'encoding': request.encoding,
         'GET': dict(request.GET),
         'POST': dict(request.POST),
         'url': request.build_absolute_uri(),
-    })
+    }
+    try:
+        if request.META.get('CONTENT_TYPE') == 'application/json':
```
kylef, line 58: I think there's a few cases where this check won't work out properly given the spec (https://tools.ietf.org/html/rfc7231#section-3.1.1.5).

1. Media Type may contain parameters, for example: `application/json; charset=utf-8`
2. [RFC 7231 3.1.1.1](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) specifies case-insensitive comparisons for the type/subtype components. I.e, `Application/JSON` should be equivilent to `application/json` for purposes of comparison here.
    > The type, subtype, and parameter name tokens are case-insensitive
3. Should this match against subtypes when there is a suffix? i.e, content-type such as `application/schema+json`

steve-nester-uk: Maybe `if request.META.get('CONTENT_TYPE').lower().startswith('application/json')` ?

bugsnag/tornado/__init__.py
```diff
@@ -1,13 +1,43 @@
...
-    def _handle_request_exception(self, exc):
+    def add_tornado_request_to_notification(self, notification):
+        if not hasattr(self, "request"):
+            return
+
+        try:
+            request_tab = {
+                'method': self.request.method,
+                'path': self.request.uri,
+                'GET': {},
```
imjoehaines, line 18: We should be able to populate this with `request.query`

```suggestion
                'GET': self.request.query,
```

tests/integrations/test_tornado.py
```diff
@@ -0,0 +1,108 @@
+import bugsnag
+import json
+from tests.utils import IntegrationTest
+from tornado.testing import AsyncHTTPTestCase
+import tests.fixtures.tornado.server
```
imjoehaines, line 5: `from tests.fixtures.tornado import server` should let you refer to just `server.make_app` on line 10

bugsnag/tornado/__init__.py
```diff
@@ -1,13 +1,43 @@
...
+            }
+            if (len(self.request.body) > 0):
+                headers = self.request.headers
+                body = self.request.body.decode('utf-8')
+                aj = 'application/json'
+                if headers.get('Content-Type').lower().startswith(aj):
+                    if request_tab["method"] == "POST":
+                        request_tab["POST"] = json.loads(body)
+                else:
+                    request_tab["POST"] = urllib.parse.parse_qs(body)
```
imjoehaines, line 32: This only works for `application/x-www-form-urlencoded` data, so I wonder if we should try using `body_arguments` instead and do nothing in case it doesn't exist — it was introduced in 2014 so should be pretty universal by now

```suggestion
                    request_tab["POST"] = self.request.body_arguments
```

bugsnag/tornado/__init__.py
```diff
@@ -1,13 +1,43 @@
...
 
+            notification.add_tab("request", request_tab)
+        except Exception:
+            pass
+
+        notification.add_tab("environment", dict(self.request.headers))
```
imjoehaines, line 38: I think this is intended to be the WSGI environment, rather than just the request headers. Something like this should work:

```suggestion
        notification.add_tab("environment", tornado.wsgi.WSGIContainer.environ(self.request))
```

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,24 @@ def add_django_request_to_notification(notification):
...
         'path': request.path,
         'encoding': request.encoding,
         'GET': dict(request.GET),
         'POST': dict(request.POST),
         'url': request.build_absolute_uri(),
-    })
+    }
+    try:
+        aj = 'application/json'
+        if request.META.get('CONTENT_TYPE').lower().startswith(aj):
```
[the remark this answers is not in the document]
steve-nester-uk: Fixed with 3977f604592167bbbf3256697467ce99adcd44f7

bugsnag/tornado/__init__.py
```diff
@@ -1,13 +1,42 @@
...
+        request_tab = {
+            'method': self.request.method,
+            'path': self.request.path,
+            'GET': urllib.parse.parse_qs(self.request.query),
+            'POST': {},
+            'url': self.request.full_url(),
+        }
+        if (len(self.request.body) > 0):
+            headers = self.request.headers
+            body = self.request.body.decode('utf-8')
```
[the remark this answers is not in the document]
steve-nester-uk: Thanks @imjoehaines, I widened the try to include all cases

bugsnag/django/__init__.py
```diff
@@ -45,14 +46,24 @@ def add_django_request_to_notification(notification):
...
         'GET': dict(request.GET),
         'POST': dict(request.POST),
         'url': request.build_absolute_uri(),
-    })
+    }
+    try:
+        aj = 'application/json'
+        if request.META.get('CONTENT_TYPE', '').lower().startswith(aj):
+            if request_tab["method"] == "POST":
+                body = request.body.decode('utf-8')
```
kattrali, line 61: In the case  that `body` can't be decoded as UTF-8, this could instead use a `replace` strategy to handle the errors and still include as much as possible in the event on  the dashboard:

```suggestion
                body = request.body.decode('utf-8', 'replace')
```

bugsnag/tornado/__init__.py
```diff
@@ -1,13 +1,42 @@
...
+            'method': self.request.method,
+            'path': self.request.path,
+            'GET': urllib.parse.parse_qs(self.request.query),
+            'POST': {},
+            'url': self.request.full_url(),
+        }
+        try:
+            if (len(self.request.body) > 0):
+                headers = self.request.headers
+                body = self.request.body.decode('utf-8')
```
kattrali, line 24: Same as above
```suggestion
                body = self.request.body.decode('utf-8', 'replace')
```

bugsnag/django/__init__.py
steve-nester-uk: Thanks @kattrali, fixed in 5db02996a3bb847eddda4c58d895f29ec3a289a4

PR #194, opened 2020-07-15
Enable session tracking by default

kattrali:
Turns on session tracking with the following changes:

* `Configuration.auto_capture_sessions` is now `True` by default
* Added constants for default endpoint values
* Attempting to send a session when `Configuration.endpoint` is customized but `Configuration.sessions_endpoint` is not issues a single warning

### DIFF
--- a/tests/test_delivery.py
+++ b/tests/test_delivery.py
@@ -1,3 +1,4 @@
+import warnings
 import sys
 
 from bugsnag import Configuration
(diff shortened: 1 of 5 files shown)


tests/test_delivery.py
```diff
@@ -44,3 +45,21 @@ def test_create_default_delivery(self):
...
+    def test_misconfigured_sessions_endpoint_sends_warning(self):
+        delivery = create_default_delivery()
+
+        with warnings.catch_warnings(record=True) as warn:
+            warnings.simplefilter("always")
+            delivery.deliver_sessions(self.config, '{"apiKey":"aaab"}')
+            self.assertEqual(1, len(warn))
+            self.assertTrue('No sessions will be sent' in str(warn[0].message))
+            delivery.deliver_sessions(self.config, '{"apiKey":"aaab"}')
+            self.assertEqual(1, len(warn))
```
imjoehaines, line 58: Is it worth asserting that nothing's actually been sent here?

PR #195, opened 2020-07-20
Add support for ASGI

kattrali:
Adds a new middleware to support reporting from ASGI-based web servers and frameworks.

## New

* `bugsnag.asgi` - Contains a `BugsnagMiddleware` class which can wrap an ASGI handler

## Fixed

* Ensured `bytes` objects are decoded prior to payload delivery

## Tests

* Added new handler tests for payload contents using [starlette](https://www.starlette.io/testclient/) as it had a nice test client


thomasleveil:
Many thanks @kattrali, this feature is very welcome :heart:

PR #197, opened 2020-07-20
Upgrade the flask example app to flask 1.1.2

kattrali:
Added a bit of code style fixes along the way 💄


example/flask/server.py
```diff
@@ -128,21 +162,22 @@ def notifywithmetadata():
             }
         },
     )
-    return ('Metadata was added to the notification, check <a href=\"bugsnag.com\">bugsnag.com</a> ' +
-        ' to view on the "Request info" and "Resolve info" tabs')
+    return ('Metadata was added to the notification, check' +
```
imjoehaines, line 165: Without a space this get smushed into the anchor tag

```suggestion
    return ('Metadata was added to the notification, check ' +
```

PR #205, opened 2020-07-30
tests(celery): add tests for the celery integration

kattrali:
* Adds a new test utility `wait_for_request` for use in non-`async` contexts
* Tests shared tasks vs app context tasks

### DIFF
--- a/tests/integrations/test_celery.py
+++ b/tests/integrations/test_celery.py
@@ -1,0 +1,186 @@
+from contextlib import contextmanager
+import pytest
+import celery
+from celery import shared_task
+from celery.signals import task_failure
+from bugsnag.celery import connect_failure_handler, failure_handler
+from tests.utils import MissingRequestError
+
+
+@pytest.fixture(scope='function')
+def celery_config():
+    return {
+        'broker_url': 'memory://',
+        'result_backend': 'rpc',
+    }
+
+
+@contextmanager
+def celery_failure_handler():
+    """
+    The bugsnag celery integration works by listening to the celery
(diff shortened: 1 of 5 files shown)


tests/integrations/test_celery.py
```diff
@@ -0,0 +1,192 @@
...
+    This context manages the signal connection and ensures that error handling
+    does not occur across separate tests.
+    """
+    connect_failure_handler()
+    try:
+        yield
+        task_failure.disconnect(failure_handler)
+    except Exception:
+        task_failure.disconnect(failure_handler)
+        raise
```
imjoehaines, line 33: Could this be simplified? I think this is equivalent

```suggestion
    try:
        yield
    finally:
        task_failure.disconnect(failure_handler)
```

tests/integrations/test_celery.py
```diff
@@ -0,0 +1,192 @@
...
+
+    celery_worker.reload()
+
+    with celery_failure_handler():
+        assert square.delay(3).get(timeout=1) == 9
+
+    try:
+        bugsnag_server.wait_for_request()
+    except MissingRequestError:
+        pass
```
imjoehaines, line 53: I'm not sure I'm convinced, but maybe this is clearer using `pytest.raises` to indicate that we expect an exception?

```suggestion
    with pytest.raises(MissingRequestError):
        bugsnag_server.wait_for_request()
```

also on line 152

tests/integrations/test_celery.py
kattrali: it is! checked by breaking some stuff. 😄

PR #208, opened 2020-08-04
Add tests against Django 3

kattrali:
## Goal

Expand the current integration testing to include Django 3, combining the existing Django test fixtures with a new set to handle changes to the app layout.

## Changeset

* There are now two django test fixtures, in `django1` (which supports django2) and `django30`. This change is in its own commit since it involves a lot of files and it should make the rest easier to review.
* There is a new integration suite in `test_django.py`
* Added [pytest-django](https://pytest-django.readthedocs.io/) to simplify migrations / setup before each test, which removed the need for separate migration and setup steps in the environment configuration
* [CI] Added [tox-factor](https://github.com/rpkilby/tox-factor) to simplify the increasingly complex build matrix. Supports running a set of tests by picking all environments with an element (e.g. `TOXFACTOR=py38 tox` will run all python 3.8 test envs


.travis.yml
```diff
@@ -10,33 +10,33 @@ matrix:
...
+    env: TOXFACTOR=py27
 
   - python: 3.3
     install:
     - travis_retry pip install coveralls 'virtualenv<16'
-    - pip install 'tox<3' 'pluggy<0.6.0' webtest
+    - pip install 'tox<3' 'pluggy<0.6.0' tox-factor webtest
     script:
-      - TOXENV=py33-test,py33-requests,py33-wsgi,py33-flask,py33-tornado tox
+      - TOXFACTOR=py33
```
imjoehaines, line 20: This is the full test script, so it isn't running tests anymore because `tox` is missing

I think the line above can be `env` rather than `script` to match the rest, but GitHub won't let me leave a suggestion on that line :s

```suggestion
      - TOXFACTOR=py33 tox
```

kattrali: Ah! nice catch.

PR #211, opened 2020-08-10
Validate and warn for unknown or incorrectly typed options

### LINKED ISSUE #193: Incorrect configuration does not generate warning or error.
### Description
Improve user experience with library by validating config instead of having a silent fault. I had to set breakpoints and step through the library to determine that this was happening. It would really help improve usability if this were more explicit :)

### Issue
https://github.com/bugsnag/bugsnag-python/blob/master/bugsnag/configuration.py#L99

You intend for Configuration.should_notify to **not** send a notification to bugsnag if no release stages are specified OR if self.notify_release_stages is not a tuple or list. The problem with this is that a user can provide incorrect config values. e.g.

### (code sample to reproduce the problem)
    bugsnag.configure(
        api_key=config.get('API_KEY'),
        release_stage='staging'
        notify_release_stages='staging'
    )

This is an invalid configuration but no exception is thrown nor warning printed. Instead the bugsnag server is not notified.
If the config is not valid, then the library should let the user know that.

### Environment

Library versions:
- python version:  python 3.7
- bugsnag version: 3.6.1

kattrali:
## Goal

When configuring the bugsnag client, emit warnings for incorrectly used configuration.

## Changeset

* [utils] Adds helpers to check if a setter function is of a provided type, otherwise issuing a warning
* [config] Adds a property for each config option, using the validation helpers to check each type
* [config] Emit a warning if an unknown configuration option is specified

## Tests

* Added new tests for each config property and setting unknown options

## Discussion

This is a larger change because its the first time all config options have been enumerated, which has the added benefit of adding a place for documentation of each one.

In a future (breaking) release:

1. the class hierarchy for `Configuration` and `RequestConfiguration` should be separated since the behavior differs completely aside from having a helper function to set multiple options at once. `Configuration` should enumerate all valid options in `configure()`, catching potential problems earlier.

1. the package as a whole can raise the minimum supported py version (from 2.6) to take advantage of the `typing` package to reduce the amount of type checking boilerplate

1. config options which are essential for the correct functioning of the library (api key, endpoint, delivery) should throw `TypeError` for incorrect types and `ValueError` for unusable values of the correct type (e.g. empty strings).

### Linked issues

Fixes #193

### DIFF
--- a/bugsnag/configuration.py
+++ b/bugsnag/configuration.py
@@ -129,7 +129,7 @@ class Configuration(_BaseConfiguration):
         """
         return self._api_key
 
-    @api_key.setter
+    @api_key.setter  # type: ignore
     @validate_required_str_setter
     def api_key(self, value):
         self._api_key = value
@@ -141,7 +141,7 @@ class Configuration(_BaseConfiguration):
         """
         return self._app_version
 
-    @app_version.setter
+    @app_version.setter  # type: ignore
     @validate_str_setter
     def app_version(self, value):
         self._app_version = value
@@ -153,7 +153,7 @@ class Configuration(_BaseConfiguration):
         """
         return self._asynchronous
 
-    @asynchronous.setter
+    @asynchronous.setter  # type: ignore
     @validate_bool_setter
     def asynchronous(self, value):
         self._asynchronous = value
@@ -166,7 +166,7 @@ class Configuration(_BaseConfiguration):
         """
         return self._auto_capture_sessions
 
-    @auto_capture_sessions.setter
+    @auto_capture_sessions.setter  # type: ignore
     @validate_bool_setter
     def auto_capture_sessions(self, value):
         self._auto_capture_sessions = value
@@ -178,7 +178,7 @@ class Configuration(_BaseConfiguration):
         """
         return self._auto_notify
 
-    @auto_notify.setter
+    @auto_notify.setter  # type: ignore
     @validate_bool_setter
     def auto_notify(self, value):
         self._auto_notify = value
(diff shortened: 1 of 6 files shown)


bugsnag/configuration.py
```diff
@@ -104,6 +95,295 @@ def __init__(self):
...
+            'ignore_classes', 'lib_root', 'notify_release_stages',
+            'params_filters', 'project_root', 'proxy_host', 'release_stage',
+            'send_code', 'session_endpoint', 'traceback_exclude_modules',
+            'use_ssl',
+        ]
+
+        for option_name in options.keys():
+            if option_name not in configurable_options:
+                message = 'received unknown configuration option "{0}"'
+                warnings.warn(message.format(option_name), RuntimeWarning)
```
imjoehaines, line 119: This is maybe a BC break — if someone was storing arbitrary data on the Configuration instance then this change would break that, so it might be safer to always do the `setattr` call

I'm happy to make this change as I think it's far more likely for this to be done in error (e.g. a typo in a config value name)

bugsnag/utils.py
```diff
@@ -233,6 +235,38 @@ def package_version(package_name):
             return None
 
 
+VALIDATION_FAILURE_FORMAT = ('{0} should be {1}, got {2}.' +
+                             ' This will be an error in a future release.')
```
imjoehaines, line 239: I think the only configuration issue that should error is a missing API key; everything else should be a warning

kattrali: Makes sense - I went with API key, endpoints, and delivery, since any of those being incorrect prevents essential usage.

PR #214, opened 2020-08-14
Add option to enable/disable sending request env

kattrali:
## Goal

It should be possible to skip collecting request environment altogether using a single option.

### Added

* `Configuration.send_environment` (bool)

### Changed

* Web framework integrations now check send_env before attaching environment tab

## Tests

* Added new tests to validate the config option
* Added new integration tests for each framework to ensure disabling send_env works

### DIFF
--- a/bugsnag/asgi.py
+++ b/bugsnag/asgi.py
@@ -100,6 +100,8 @@ class BugsnagMiddleware:
             request['url'] = parse_url(request, server)
 
             event.add_tab("request", request)
+            if bugsnag.configure().send_environment:
+                event.add_tab("environment", scope)
 
         stack.before_notify(add_request_info)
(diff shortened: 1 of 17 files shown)


bugsnag/asgi.py
```diff
@@ -100,6 +100,8 @@ def add_request_info(event):
             request['url'] = parse_url(request, server)
 
             event.add_tab("request", request)
+            if bugsnag.configure().send_environment:
+                event.add_tab("environment", scope)
```
imjoehaines, line 104: Should we add that we now collect the environment in ASGI apps to the changelog too?

kattrali: oh yeah, good call!

PR #217, opened 2020-08-18
examples: Add django 3.1 example project

kattrali:
## Goal

Add some latest and greatest usage examples


example/django31/todo/wsgi.py
```diff
@@ -1,14 +1,16 @@
 """
-WSGI config for bugsnag_demo project.
+WSGI config for todo project.
 
 It exposes the WSGI callable as a module-level variable named ``application``.
 
 For more information on this file, see
-https://docs.djangoproject.com/en/1.6/howto/deployment/wsgi/
+https://docs.djangoproject.com/en/3.0/howto/deployment/wsgi/
```
imjoehaines, line 7: Should this go to the 3.1 docs?

```suggestion
https://docs.djangoproject.com/en/3.1/howto/deployment/wsgi/
```

kattrali: hmm... yeah 👀 

I went through and updated the other urls in the app as well.

PR #219, opened 2020-08-20
Add platform to config runtime_versions

tschoonj:
## Goal

The platform that the app runs on may be useful information for debugging.

## Design

`Configuration`'s `runtime_versions` seemed like a good place to attach this piece of information.

## Changeset

The `runtime_versions` attribute of `Configuration` instances now includes the platform key in its dict.

## Testing

A unittest was added to confirm it works

### DIFF
--- a/bugsnag/configuration.py
+++ b/bugsnag/configuration.py
@@ -97,7 +97,10 @@ class Configuration(_BaseConfiguration):
         else:
             self.hostname = None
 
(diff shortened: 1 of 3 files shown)


steve-nester-uk:
Thanks for the PR, the runtimeVersion data should be more structured than just a human-readable string.  So we will not proceed with this particular solution.  One option for adding this information would be to use custom metadata to pass to Bugsnag in an error report.

PR #223, opened 2020-09-01
[v4] raise minimum supported py version to 3.5

kattrali:
First changeset leading to the 4.0 release - raises the minimum supported version to python 3.5, which is the oldest version not currently at end-of-life.

* updated the test matrix
* added the versioning requirements to the package
* added new entries to the changelog and upgrading guide
* removed 2.7-specific packaging setup

This is the base for pretty much everything to follow: improved asyncio integration, removing now-unneeded code/deprecations, and more


setup.py
```diff
@@ -21,6 +21,7 @@
     author_email='simon@bugsnag.com',
     url='https://bugsnag.com/',
     license='MIT',
+    python_requires='>=3.4.*, <4',
```
tomlongridge, line 24: Shouldn't this be >=3.5?

```suggestion
    python_requires='>=3.5.*, <4',
```

kattrali: yup! looked at it so many times and then missed it.

tomlongridge (approved):

PR #233, opened 2020-09-09
style: designate public vs internal APIs with __all__ 

kattrali:
This change doesn't prevent using any particular piece (if you know what you're looking for) but [serves as an indicator](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces) for what can change without notice in the future.


imjoehaines (approved):
Do we need a changelog/upgrading entry for this, as it changes `import *`?

kattrali:
yeah! added a general note about import usage to address that.

PR #235, opened 2020-09-21
refactor(legacy): remove unused bugsnag.send_sessions

kattrali:
## Goal

This helper is unused, and accessing the session tracker directly could be done if needed


imjoehaines (approved):
👋

PR #239, opened 2020-09-28
tests(celery): test against v4 and v5

kattrali:
Celery 5.0 removes the pytest fixtures and instead puts them in a
separate pytest-celery package.


imjoehaines (approved):
~Coveralls is upset but~ looks good to me!

PR #240, opened 2020-09-28
rename 'meta_data'

kattrali:
## Goal

Rename `meta_data` to metadata, which should be more in line with the spec and the dictionary.

## Changeset

* Updated `Event.meta_data` and `RequestConfiguration.meta_data`, adding deprecation warnings
* Updated every use of `meta_data` to `metadata` in the library. This is in the latter commit to make it easier to review.

## Testing

* Tested that the library still functioned as expected after the initial change (but still using `meta_data` everywhere). The only test breakages were related to the tests which verified the number of warnings sent.
* Manually sent events to verify payload contents
* Added new tests for the warnings themselves

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ Please upgrade.
   `bugsnag.event.Event` to better align with Bugsnag libraries on other
   platforms. The `Notification` class is functionally equivalent and will be
   removed in a future release.
+* `Event.meta_data` has been renamed to `Event.metadata`
+* `RequestConfiguration.meta_data` has been renamed to
+  `RequestConfiguration.metadata`
 
 ### Enhancements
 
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ Please upgrade.
   `bugsnag.event.Event` to better align with Bugsnag libraries on other
   platforms. The `Notification` class is functionally equivalent and will be
   removed in a future release.
+* `Event.meta_data` has been renamed to `metadata`
+* `RequestConfiguration.meta_data` has been renamed to `metadata`
```
imjoehaines, line 38: This might be a little confusing with just `metadata`

```suggestion
* `Event.meta_data` has been renamed to `Event.metadata`
* `RequestConfiguration.meta_data` has been renamed to `RequestConfiguration.metadata`
```

imjoehaines (approved):

PR #248, opened 2020-12-03
ci: notify coveralls when parallel jobs complete

kattrali:
Avoid situation where first job to complete is deemed the entire build's coverage.

[Found some useful docs on the topic](https://github.com/coveralls-clients/coveralls-python/blob/458e22defbdeaf86458cccf315eee5000bf97689/docs/usage/configuration.rst#github-actions-gotcha)


imjoehaines (approved):
𝓷𝓲𝓬𝓮 😎

PR #249, opened 2020-12-07
Support for level parameter in constructor

ForroKulcs:
## Goal

logging.Handler supports level parameter. This change is to enable forwarding this value.

## Design

level is introduced as first parameter as it is also the first parameter of logging.Handler

## Changeset

constructor declaration was changed

## Testing

manual local test was used

### DIFF
--- a/bugsnag/handlers.py
+++ b/bugsnag/handlers.py
@@ -9,11 +9,11 @@ __all__ = ('BugsnagHandler',)
 
 
 class BugsnagHandler(logging.Handler, object):
-    def __init__(self, client=None, extra_fields=None):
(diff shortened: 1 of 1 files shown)


johnkiely1:
Hi @ForroKulcs .

Thanks for the PR, this sounds like a good idea. We will review in due course.

SB-PawanN:
Thank you for your contribution. This is included in 416(https://github.com/bugsnag/bugsnag-python/pull/416)

SB-jigneshR:
This is included in 416(https://github.com/bugsnag/bugsnag-python/pull/416)

PR #263, opened 2021-05-11
Move internal Bugsnag logger to the Configuration class

imjoehaines:
## Goal

We have a logger that's used internally when various error conditions happen, e.g. when delivery fails. Currently this is exposed as `bugsnag.logger`

This has been moved to the Configuration class, in order to be more inline with notifiers for other platforms and to allow for logging to use configuration options in future

A logger can be passed to the Configuration constructor and `configure` method:

```python
config = Configuration(logger=some_logger)
config.configure(logger=some_other_logger)
```

Logging can be disabled by passing `None` instead of a logger instance

Any value other than a `logging.Logger` or `None` will raise a warning and use the default logger

`bugsnag.logger` is still available for backwards compatibility using the global configuration instance's logger

### DIFF
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -1,6 +1,9 @@
 import os
 import socket
+import logging
 import unittest
+from unittest.mock import patch
+from io import StringIO
 
 from bugsnag.configuration import Configuration
 from bugsnag.middleware import DefaultMiddleware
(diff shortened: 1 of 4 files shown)


tests/test_configuration.py
```diff
@@ -362,3 +365,138 @@ def test_validate_unknown_config_option(self):
...
+
+    @patch('sys.stderr', new_callable=StringIO)
+    def test_default_logger_is_used_if_invalid_logger_set(self, mock_stderr):
+        with pytest.warns(RuntimeWarning) as warnings:
+            config = Configuration(logger=123)
+
+        expected_message = 'logger should be logging.Logger, got int'
+
+        assert len(warnings) == 1
+        assert warnings[0].message.args[0] == expected_message
```
kattrali, line 471: Shouldn't this be inside the context at line 465 to access `warnings`? Seems out that its accessible outside of that scope.

imjoehaines: `with` doesn't create a scope so it's accessible here; this is straight out of the docs: https://docs.pytest.org/en/stable/warnings.html#warns

kattrali: ah! 💡

PR #264, opened 2021-05-12
Use the Configuration logger instead of `bugsnag.logger`

imjoehaines:
## Goal

Following up from #263, this PR replaces calls to `bugsnag.logger` with the Configuration's logger

There are a couple of places where this wasn't totally trivial; the `SanitizingJSONEncoder` and the Django integration so these are done in separate commits to be easier to review

### DIFF
--- a/bugsnag/django/__init__.py
+++ b/bugsnag/django/__init__.py
@@ -44,7 +44,7 @@ def add_django_request_to_notification(event):
                 username = str(request.user.get_username())
                 event.set_user(id=username, email=email, name=name)
             except Exception:
-                bugsnag.logger.exception('Could not get user data')
+                event.config.logger.exception('Could not get user data')
     else:
         event.set_user(id=request.META['REMOTE_ADDR'])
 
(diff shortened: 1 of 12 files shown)


bugsnag/django/__init__.py
```diff
@@ -72,36 +72,48 @@ def add_django_request_to_notification(event):
...
     middleware.before_notify(add_django_request_to_notification)
 
-    bugsnag.configure().runtime_versions['django'] = django.__version__
+    config.runtime_versions['django'] = django.__version__
+
+    request_started.connect(__track_session)
+    got_request_exception.connect(
+        __handle_request_exception(config.logger),
+        weak=False
+    )
```
imjoehaines, line 94: Django stores signal handlers as a weak reference by default, which doesn't work when this is using a closure (to have access to the logger) as it gets garbage collected. Setting `weak=False` makes it a strong reference so this works

image

imjoehaines:
The license audit is currently broken for unrelated reasons; I'll fix it in a separate PR

kattrali (approved):

PR #267, opened 2021-05-19
Make sure you filter values that are not strings

Bachmann1234:
## Goal

I want to be able to filter out sensitive info even if they request payload is coming to you as bytes

## Design

Because it kept the literal changes minimal 

## Changeset

I copy the filters to encoded versions of those filters and then I handle bytes in the sensitive encoder

## Testing

I added a test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ htmlcov
 #IDE files
 .vscode
 my_env
+venv
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+## TBD
+
+### Bug fixes
+
+* Apply filtering with `params_filters` to bytes, not just strings
+  [Matt Bachmann](https://github.com/Bachmann1234)
+  [#267](https://github.com/bugsnag/bugsnag-python/pull/267)
+
 ## 4.0.3 (2021-04-28)
 
 ### Bug fixes
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -35,7 +35,7 @@ class SanitizingJSONEncoder(JSONEncoder):
 
     def __init__(self, keyword_filters=None, **kwargs):
         self.filters = list(map(str.lower, keyword_filters or []))
-        self.bytes_filters = [x.lower().encode('utf-8') for x in keyword_filters or []]
+        self.bytes_filters = [x.encode('utf-8') for x in self.filters]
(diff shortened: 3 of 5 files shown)


johnkiely1:
Hi @Bachmann1234, 

Thanks for the PR. 
We will look to review as soon as priorities allow.

Bachmann1234:
@johnkiely1 thank you! 

I just wanna make it clear I see this as a bug fix not a feature request. While I am not 100% sure the mechanism the request data being sent to bugsnag was bytes and sensitive auth data was not being filtered out (luckily I noticed this during my local testing) 

A bit about the stack

python 3.9
web framework is starlette (asgi)
And I am using a logging framework 'structlog` to provide json structured logs.

I suspect its something about how starlette stores the request data but I am not entirely sure. I just discovered bytes were being sent when I threw a debugger in the library

mattdyoung:
Thanks for the additional context @Bachmann1234 - we've assigned this to an upcoming sprint for review.

Bachmann1234:
@mattdyoung Thank you! Let me know if you need anything or have changes I should make.

ALso I wanted to rule out any weirdness in our app so here is a minimal example

```
import os
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.routing import Route

import bugsnag
from bugsnag.asgi import BugsnagMiddleware


async def homepage(request):
    raise ValueError("OMG WHAT")
    return JSONResponse({"hello": "world"})


bugsnag.configure(api_key=os.environ["BUGSNAG_KEY"])
app = BugsnagMiddleware(
    Starlette(
        debug=True,
        routes=[
            Route("/", homepage),
        ],
    )
)
```

python version
```
Python 3.9.4
```

output of pip freeze
```
bugsnag==4.0.3
click==7.1.2
h11==0.12.0
starlette==0.14.2
uvicorn==0.13.4
WebOb==1.8.7
```

start the server
```
uvicorn start:app
```

example request
`curl localhost:8000 -H 'Authorization: 123'`

Result
Screen Shot 2021-05-21 at 1 58 55 PM

philipbjorge:
@Bachmann1234 - Really appreciate you putting this PR up here.

@mattdyoung -
We're working on a FastAPI project and noticed that Bugsnag was leaking authorization credentials during development.
It seems somewhat likely to me that there are other teams leaking similar sensitive data right now because of this bug.

imjoehaines (approved):
Thanks for the PR @Bachmann1234!

PR #268, opened 2021-05-19
Add Breadcrumb class and BreadcrumbType enum

imjoehaines:
## Goal

Adds the basic building blocks of breadcrumbs — the BreadcrumbType enum and Breadcrumb class

At the moment these aren't connected to anything, but will be used in future PRs

## Testing

Unit tests have been added, though these additions don't really have any behaviour by themselves (more useful tests depend on future additions)

### DIFF
--- a/bugsnag/__init__.py
+++ b/bugsnag/__init__.py
@@ -2,6 +2,7 @@ from bugsnag.configuration import Configuration, RequestConfiguration
 from bugsnag.notification import Notification
 from bugsnag.event import Event
 from bugsnag.client import Client
+from bugsnag.breadcrumbs import BreadcrumbType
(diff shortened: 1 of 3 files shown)


bugsnag/__init__.py
```diff
@@ -2,6 +2,7 @@
 from bugsnag.notification import Notification
 from bugsnag.event import Event
 from bugsnag.client import Client
+from bugsnag.breadcrumbs import BreadcrumbType, Breadcrumb
```
imjoehaines, line 5: I'm not sure whether we should be exposing these classes, but I think this is useful for type annotations

kattrali (approved):

PR #269, opened 2021-05-20
Add Breadcrumbs class to manage breadcrumbs

imjoehaines:
## Goal

This PR adds the `Breadcrumbs` class, which manages a `ContextVar` containing a `deque` of breadcrumbs. This allows many breadcrumbs to be added, but only the most recent 𝑥 breadcrumbs will be kept

With some async APIs the ContextVar has to be explicitly copied, otherwise it can be shared (e.g. in the async unit test) but in others this is unnecessary (e.g. `asyncio` servers). In a future PR, this will be handled much like per-request data is currently — the `create_copy_for_context` method will be called in various integration middlewares

### DIFF
--- a/bugsnag/breadcrumbs.py
+++ b/bugsnag/breadcrumbs.py
@@ -1,6 +1,26 @@
 from enum import Enum, unique
-from typing import Any, Dict, Union
+from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
+from collections import deque
+
 from bugsnag.utils import FilterDict
+
+# Deque is not present in 'typing' until 3.5.4, so we can't use it directly
+if TYPE_CHECKING:
+    from typing import Deque
+
+# The _breadcrumbs context var contains None or a deque of Breadcrumb instances
+try:
+    from contextvars import ContextVar
+    _breadcrumbs = ContextVar(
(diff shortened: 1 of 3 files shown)


bugsnag/breadcrumbs.py
```diff
@@ -40,3 +60,47 @@ def to_dict(self) -> Dict[str, Union[str, FilterDict]]:
...
+        _breadcrumbs.set(new_breadcrumbs)
+        self._max_breadcrumbs = new_max
+
+    # Create a copy of the current list of breadcrumbs for this context
+    def create_copy_for_context(self) -> None:
+        # Resizing will create a new deque and store it in the ContextVar,
+        # which will give the current context a new copy of the list
+        self.resize(self._max_breadcrumbs)
+
+    def to_list(self) -> List[Breadcrumb]:
```
kattrali, line 91: This class could instead implement `__iter__` and `__next__`, if you want `list(breadcrumbs)` to work

kattrali (approved):

bugsnag/breadcrumbs.py
kattrali: Yeah that's more than reasonable. I figured you might have gone down that route already but if not, I'd check. :+1:

PR #271, opened 2021-05-24
Add to_rfc3339 for formatting datetimes with millisecond precision

imjoehaines:
## Goal

For breadcrumbs to display nicely in the dashboard, it's important we report them with millisecond precision. Python 3.6 and above can do this natively with `datetime.isoformat(timespec='milliseconds')` but on 3.5 the only options are seconds precision or microseconds precision. Timestamps with microseconds appear to get dropped by the API, so we need to implement a fallback for Python 3.5

This PR adds a `to_rfc3339` function to `bugsnag.utils` which formats a datetime object to match to_rfc3339/iso8601 with milliseconds precision, e.g. `1900-01-02T03:04:05.678+12:00`

### DIFF
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -6,6 +6,7 @@ from typing import Tuple, Optional
 import warnings
 import copy
 import logging
+from datetime import datetime, timedelta
 
 MAX_PAYLOAD_LENGTH = 128 * 1024
 MAX_STRING_LENGTH = 1024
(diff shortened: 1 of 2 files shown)


bugsnag/utils.py
```diff
@@ -316,3 +317,48 @@ def get(self):
...
+            sign = '-'
+            utc_offset = -utc_offset
+
+        hours_offset, minutes = divmod(utc_offset, timedelta(hours=1))
+        minutes_offset, seconds = divmod(minutes, timedelta(minutes=1))
+
+        return '{}{:02}:{:02}'.format(sign, hours_offset, minutes_offset)
+
+    def to_rfc3339(dt: datetime) -> str:
+        return '{:04}-{:02}-{:02}T{:02}:{:02}:{:02}.{:03}{}'.format(
```
kattrali, line 355: The stdlib implementation casts each component to int (`{:02d}`, etc), is that necessary here?

imjoehaines: `d` is the default for integers, but I guess it's possibly a little safer to specify it explicitly

kattrali (approved):

PR #272, opened 2021-05-24
Add `Client.leave_breadcrumb`

imjoehaines:
## Goal

Adds `leave_breadcrumb` to the Client class, which calls `on_breadcrumb` callbacks and attaches the new breadcrumb to the list

`on_breadcrumb` callbacks can "cancel" the breadcrumb by returning `False`. Any exceptions raised in callbacks will be logged but won't prevent the breadcrumb from being recorded

A `leave_breadcrumb` function has also been added to `legacy.py`, which forwards the call on to the "default client"

### DIFF
--- a/bugsnag/client.py
+++ b/bugsnag/client.py
@@ -1,14 +1,16 @@
 import sys
 import threading
 
+from datetime import datetime, timezone
 from functools import wraps
-from typing import Union, Tuple, Callable, Optional, List, Type
+from typing import Union, Tuple, Callable, Optional, List, Type, Dict, Any
 
-from bugsnag.breadcrumbs import Breadcrumb
+from bugsnag.breadcrumbs import Breadcrumb, BreadcrumbType
 from bugsnag.configuration import Configuration, RequestConfiguration
 from bugsnag.event import Event
 from bugsnag.handlers import BugsnagHandler
 from bugsnag.sessiontracker import SessionTracker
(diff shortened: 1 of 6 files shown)


bugsnag/client.py
```diff
@@ -208,6 +210,41 @@ def add_on_breadcrumb(self, on_breadcrumb) -> None:
...
+        metadata: Dict[str, Any] = {},
+        type: BreadcrumbType = BreadcrumbType.MANUAL
+    ) -> None:
+        # Don't create breadcrumbs if the max_breadcrumbs is 0 as they would
+        # be immediately discarded anyway
+        if self.configuration.max_breadcrumbs == 0:
+            return
+
+        # Default the type to "manual" if it's not a valid BreadcrumbType
+        if not isinstance(type, BreadcrumbType):
```
kattrali, line 225: Is it reasonable to coerce string values to breadcrumb types here, to make the invocation more concise? (Did we have this conversation already, possible deja vu)

```py
Bugsnag.leave_breadcrumb("did the thing", {"button": "B"}, "navigation")
```
?

kattrali (approved):

bugsnag/client.py
imjoehaines: Yeah, that makes sense to me

kattrali (approved):

PR #274, opened 2021-05-26
Automatically reset Breadcrumbs between tests

imjoehaines:
## Goal

'conftest.py' will be executed by pytest and autouse fixtures will apply to every test in the module. Tests that interact with breadcrumbs no longer need to individually reset the ContextVar between tests

See https://docs.pytest.org/en/6.2.x/example/simple.html#package-directory-level-fixtures-setups

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,9 @@
+import pytest
+from bugsnag import Breadcrumbs
+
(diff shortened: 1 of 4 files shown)


kattrali (approved):
Ah, that's much nicer.

PR #278, opened 2021-06-08
Ensure breadcrumb metadata is a dict

imjoehaines:
## Goal

This is important because:

- metadata should always be a key/value mapping
- we convert the dict into a `FilterDict` which applies filters to the metadata. If we accept anything other than dict, we open the possibility of filtering not being applied

Without this change, an exception will be raised in the `breadcrumb.to_dict` method for types that are not iterable and possibly undefined behaviour occurs for iterable types

### DIFF
--- a/bugsnag/client.py
+++ b/bugsnag/client.py
@@ -1,5 +1,7 @@
+import builtins
 import sys
 import threading
+import warnings
 
 from datetime import datetime, timezone
 from functools import wraps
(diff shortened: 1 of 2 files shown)


bugsnag/client.py
```diff
@@ -226,6 +228,15 @@ def leave_breadcrumb(
...
             type = BreadcrumbType.from_string(type)
 
+        if not isinstance(metadata, dict):
+            warning_message = 'breadcrumb metadata must be a dict, got {}'
+            warnings.warn(
+                warning_message.format(builtins.type(metadata).__name__),
+                RuntimeWarning
+            )
+
+            metadata = {}
```
imjoehaines, line 238: This matches the behaviour of JavaScript when something invalid is passed as breadcrumb metadata. Not sure if it's worth keeping the value (e.g. `metadata = {'?': metadata}`)

kattrali (approved):

bugsnag/client.py
kattrali: I think dropping it is reasonable (and might even be specified somewhere?)

PR #287, opened 2021-06-18
Avoid capturing query parameters in referer header

imjoehaines:
## Goal

Currently we use the raw `Referer` header as the `from` value in navigation breadcrumbs. This has the potential to leak sensitive information, e.g. if a query parameter contains a password

This PR adds `bugsnag.utils.sanitize_url`, which parses the given URL and returns the scheme, domain & path, and uses it on the `Referer` header when recording navigation breadcrumbs

### DIFF
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -2,7 +2,7 @@ from functools import wraps, partial
 import inspect
 from json import JSONEncoder
 from threading import local as threadlocal
-from typing import Tuple, Optional, Union
+from typing import Tuple, Optional
 import warnings
 import copy
 import logging
@@ -320,7 +320,7 @@ class ThreadContextVar:
         setattr(ThreadContextVar.local_context(), self.name, new_value)
 
 
-def sanitize_url(url_to_sanitize: str) -> Union[str, None]:
+def sanitize_url(url_to_sanitize: str) -> Optional[str]:
     try:
         parsed = urlparse(url_to_sanitize)
     except Exception:
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -2,12 +2,12 @@ from functools import wraps, partial
 import inspect
 from json import JSONEncoder
 from threading import local as threadlocal
-from typing import Tuple, Optional
+from typing import AnyStr, Tuple, Optional
 import warnings
 import copy
 import logging
 from datetime import datetime, timedelta
-from urllib.parse import urlparse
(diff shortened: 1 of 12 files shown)


bugsnag/utils.py
```diff
@@ -319,6 +320,24 @@ def set(self, new_value):
         setattr(ThreadContextVar.local_context(), self.name, new_value)
 
 
+def sanitize_url(url_to_sanitize: str) -> Union[str, None]:
```
kylef, line 323:
```suggestion
def sanitize_url(url_to_sanitize: str) -> Optional[str]:
```

imjoehaines: Good catch, thanks!

bugsnag/utils.py
```diff
@@ -319,6 +320,24 @@ def set(self, new_value):
...
+    # if the "netloc" is empty then the URL is malformed in some way - we
+    # always expect to be able to parse this from url_to_sanitize
+    if parsed.netloc == '':
+        return None
+
+    return "{}://{}/{}".format(
+        parsed.scheme,
+        parsed.netloc,
+        parsed.path.lstrip('/')
+    )
```
kattrali, line 338: - Is the `fragment` worth sending? 
- Potentially could use `urlunsplit` here to do the reassembly.

imjoehaines: I don't think `fragment` will ever be present as browsers don't include it in URLs

Not sure if there's an advantage to `urlunsplit`; it's [basically doing the same thing](https://github.com/python/cpython/blob/d597fdc5fd0e8aa73a783fea27287db669950c15/Lib/urllib/parse.py#L495-L512) but we'd have to only pass in the bits we want to keep

kattrali: > I don't think fragment will ever be present as browsers don't include it in URLs

Ah right, makes sense. Even in the [spec](https://datatracker.ietf.org/doc/html/rfc1945#section-10.13) it says:
> If a partial URI is given, it should be interpreted relative to the
   Request-URI. The URI must not include a fragment.

That first bit raises a question though - maybe there are valid cases for `Referer` which will not include `netloc`? unsplit would handle that case.

```py
>>> from urllib.parse import urlunsplit
>>> urlunsplit(('', '', '/hi', '', ''))
'/hi'
```

imjoehaines: Yeah, I guess if you're referring to a different page then you might use a relative URL for that

kattrali (approved):

PR #302, opened 2021-12-21
Add tests against Django 4

imjoehaines:
## Goal

Adds tests against Django 4. These pass without any code changes to Bugsnag itself


tests/fixtures/django4/todo/settings.py
```diff
@@ -0,0 +1,129 @@
...
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = 'django-insecure-z++&i7gx^m*qqq_1(xjqo=7%gg)^&fcq6fa3gen+10(zn5756z'
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = False
+
+ALLOWED_HOSTS = ['localhost', '0.0.0.0']
+BUGSNAG = {
+    'api_key': 'a05afff2bd2ffaf0ab0f52715bbdcffd',
+    'project_root': str(BASE_DIR),
```
imjoehaines, line 32: This is a small potential gotcha — Django is using a `Path` object from `pathlib` now, but we don't handle that as the `project_root`

I've raised a task to look at supporting this separately, as it's not really related to Django itself

kattrali (approved):

PR #305, opened 2022-01-25
Add test coverage for `traceback` Event option

imjoehaines:
## Goal

Adds a test case covering the `traceback` Event option; this allows you to pass a traceback for the event to use instead of using the one from `sys.exc_info`

This is kind of covered by existing tests, but they also pass when using `exception.__traceback__` instead of the `traceback` option because they use the same exception object to create both the Event and traceback

This new test _only_ works with the traceback option, so should help prevent regressions when making changes to this area in future


kattrali (approved):
clever :thinking:

PR #299, opened 2021-11-26
Allow multiple exceptions

drestrepom:
## Goal
- bugsnag-python only throws the first exception in the exception chain
- full information is required in bugsnag app
- there are some errors are full traceability
- there are some errors that do not contain full traceability

## Design

- bugsnag (web) supports a full stacktrace (chain of exceptions)
- get the exceptions chain based on the main exception

## Changeset

- send multiple exceptions to bugsnag
### Before

image

### After

image

### StackTrace

```
Traceback (most recent call last):
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/config/__init__.py", line 85, in load
    checks=load_checks(config),
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/config/__init__.py", line 19, in load_checks
    {core_model.FindingEnum[finding] for finding in config.pop("checks")}
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/config/__init__.py", line 19, in <setcomp>
    {core_model.FindingEnum[finding] for finding in config.pop("checks")}
  File "/nix/store/7ssj1jid9dqikk513rvg340kpbxrv0p0-python3-3.8.11/lib/python3.8/enum.py", line 387, in __getitem__
    return cls._member_map_[name]
KeyError: 'F074'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/core/scan.py", line 172, in main
    CTX.config = load(group, config)
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/config/__init__.py", line 112, in load
    raise confuse.ConfigError(f"Key: {exc.args[0]} is required")
confuse.exceptions.ConfigError: Key: F074 is required

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/utils/function.py", line 117, in wrapper
    return await function(*args, **kwargs)
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/cli/__init__.py", line 302, in cli_scan_wrapped
    success: bool = await core.scan.main(
  File "/nix/store/10qqx0mn7x11as8x4lydq16zxbxzd7fs-skims/core/scan.py", line 181, in main
[...]

PR #306, opened 2022-01-25
Add initial tests for caused by exceptions

imjoehaines:
## Goal

This PR adds some initial tests for caused by exceptions (#299)

This covers implicit (`exception.__context__`) and explicit (`exception.__cause__`) chaining, but does not cover the `raise exception from None` form[^1] — that requires some additional code changes, which will be done in a separate PR

I've added fixtures for these tests as this lets us assert against the stacktrace more easily

[^1]: The [Python `raise` docs](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) have a good explanation of each type of caused by exception

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -6,6 +6,7 @@ import inspect
 import logging
 from datetime import datetime, timedelta, timezone
 from unittest.mock import Mock, ANY
+from tests import fixtures
 
 from bugsnag import (
     Client,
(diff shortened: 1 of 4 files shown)


tests/test_client.py
```diff
@@ -928,6 +929,190 @@ def test_asynchronous_delivery_failures_are_logged(self):
...
+
+        assert len(exceptions) == 3
+
+        assert exceptions[0]['message'] == 'a'
+        assert exceptions[0]['errorClass'] == 'NameError'
+
+        # TODO: this stacktrace is generated using 'sys.exc_info()', so starts
+        #       where we construct the Event
+        #       switching to 'exception.__traceback__' should improve this as
+        #       it will use the traceback from 'exception_with_explicit_cause'
```
imjoehaines, line 948: Specifically I'm talking about changing this line in the Event class:

https://github.com/bugsnag/bugsnag-python/blob/67311be1b72c4c5023f2d44efe2a63b0e2f68dd8/bugsnag/event.py#L84

I think we're using `sys.exc_info` for historical reasons, as Python 3.0 added `exception.__traceback__` but I've not really investigated this yet. I'll look into this separately as it's not really related to caused by exceptions

kattrali: almost :100:% sure that your hunch is correct

kattrali (approved):

PR #308, opened 2022-02-10
Fix backwards compatibility breaks in caused by exceptions implementation

imjoehaines:
## Goal

This PR fixes a couple of BC breaks in the caused by exceptions implementation (https://github.com/bugsnag/bugsnag-python/pull/299):

- the `Event.stacktrace` property has been restored
- mutating or reassigning `Event.stacktrace` will take affect in the final payload
- reassigning the `Event.exception` property will take affect in the final payload

Mutating `Event.exception` will _mostly_ take affect — you can change the message but not the class name. That seems like a very unusual use-case though!

### DIFF
--- a/bugsnag/event.py
+++ b/bugsnag/event.py
@@ -1,5 +1,4 @@
 from typing import Any, Dict, Optional, List
-from types import TracebackType
 import linecache
 import logging
 import os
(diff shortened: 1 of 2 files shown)


bugsnag/event.py
```diff
@@ -179,13 +201,17 @@ def _generate_exceptions(
...
             )
 
         return trace_exceptions
 
-    def _generate_stacktrace(self, tb, source_func=None):
+    def _generate_stacktrace(
+        self,
+        tb,
+        source_func=None
+    ) -> List[Dict[str, Any]]:
```
imjoehaines, line 214: Adding this return type caused mypy to start complaining about this method

PR #309, opened 2022-02-10
Refactor caused by exceptions to match other platforms

imjoehaines:
## Goal

This PR refactors our implementation of caused by exceptions to match other platforms — the list of exceptions now lives on `Event.errors` and the original exception instance is now `Event.original_error`. For example, this matches [JavaScript](https://docs.bugsnag.com/platforms/javascript/customizing-error-reports/#errors) & [Ruby](https://docs.bugsnag.com/platforms/ruby/other/customizing-error-reports/#errors)

`Event.exception` and `Event.stacktrace` have been deprecated as they are somewhat ambiguous given there can be multiple exceptions with their own stacktraces

I've also fixed up our code to avoid triggering these deprecations internally and tidied up some tests that were leaking warnings

### DIFF
--- a/bugsnag/error.py
+++ b/bugsnag/error.py
@@ -26,5 +26,6 @@ class Error:
         return {
             'errorClass': self.error_class,
             'message': self.error_message,
-            'stacktrace': self.stacktrace
+            'stacktrace': self.stacktrace,
+            'type': self.type,
         }
--- a/bugsnag/error.py
+++ b/bugsnag/error.py
@@ -1,0 +1,30 @@
+from typing import Any, Dict, List
+
+__all__ = ('Error',)
+
+
+class Error:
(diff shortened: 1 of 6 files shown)


bugsnag/error.py
```diff
@@ -0,0 +1,30 @@
...
+        self.error_message = error_message
+        self.stacktrace = stacktrace
+        self.type = 'python'
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            'errorClass': self.error_class,
+            'message': self.error_message,
+            'stacktrace': self.stacktrace
+        }
```
kattrali, line 30: Shouldn't `type` be included?

imjoehaines: Yes!

PR #310, opened 2022-02-10
Add tests for caused by exceptions using the capture decorator/context manager

imjoehaines:
## Goal

Adds coverage of the capture decorator/context manager with caused by exceptions. These tests are essentially the same as those that manually call `notify`, except the first exception's stacktrace is improved due to the `traceback` & `source_func` parameters that `capture` passes to the Event

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1033,6 +1033,73 @@ class ClientTest(IntegrationTest):
             }
         ]
 
+    def test_chained_exceptions_with_explicit_cause_using_capture_cm(self):
+        try:
+            with self.client.capture():
+                fixtures.raise_exception_with_explicit_cause()
+        except Exception:
+            pass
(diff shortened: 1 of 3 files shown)


tests/test_client.py
```diff
@@ -1151,6 +1285,57 @@ def test_chained_exceptions_with_no_cause(self):
         assert exceptions[0]['stacktrace'][0]['method'] == \
             'test_chained_exceptions_with_no_cause'
 
+    def test_chained_exceptions_with_no_cause_using_capture_cm(self):
```
kattrali, line 1288: are the names of this test and the one below it swapped?

imjoehaines: yeah, not sure how that happened! 🤦

kattrali (approved):

PR #326, opened 2022-11-02
Prevent duplicate events in Django integration

imjoehaines:
## Goal

We use both a middleware and signal handler to catch exceptions in Django applications, but both can be called for the same exception which causes a duplicate event to be reported

Using the new `skip_bugsnag` attr, we can prevent this by stopping the signal handler from notifying any exception that has already been handled by the middleware. The signal handler is still necessary as some exception happen outside of the middleware stack

## Changeset

- The signal handler is no longer marked as `weak=True`. This was causing issues when asserting on the number of events delivered in the tests because it was persisting between tests. This wasn't reproducible in a real app though as a real app won't start Django multiple times
- Our Django middleware now sets `skip_bugsnag` on exceptions it handles. The middleware runs before the signal handler so this prevents the signal handler from also notifying of the same exceptions

## Testing

- FakeBugsnagServer now clears received events between tests to allow asserting on only the number of requests delivered in an individual test
- FakeBugsnagServer always waits for the entire duration of its `timeout` so that excess requests have a chance to be delivered. This causes tests to run longer, but it's not a huge impact (total time is still ~1-3 mins)
- The Django tests now assert that only 1 event is reported. Doing this on `next` fails as both the signal handler and middleware report the same exception in some tests (even with the other testing changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Changelog
 * Allow preventing an exception from being reported to Bugsnag by setting the `skip_bugsnag` attr to `True`
   [#325](https://github.com/bugsnag/bugsnag-python/pull/325)
 
+* Prevent duplicate events from being notified in the Django integration
+  [#326](https://github.com/bugsnag/bugsnag-python/pull/326)
+
 ## v4.2.1 (2022-05-16)
 
 ### Bug fixes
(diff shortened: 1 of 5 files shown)


kattrali (approved):
This is great!

PR #328, opened 2022-12-15
Updating the requirements.txt file + small change in urls.py

hannah-smartbear:
## Goal

To make the app run, since the requirements were outdated

## Design

checked the latest version of all the requirements and updated them

## Changeset

The requirements and one line in the urls.py file

## Testing

I ran the project using the new requirements file to confirm that it ran.


example/celery+django/requirements.txt
```diff
@@ -1,11 +1,11 @@
-amqp==2.2.1
-billiard==3.5.0.3
-bugsnag==3.1.1
```
luke-belton, line 3: Do we want to pin the bugsnag version? Feels like the example should get the latest features!

luke-belton (approved):
LGTM.

@imjoehaines there are some failing tests (looks like the required Python versions can't be installed??). Please could you take a look at this as I don't think Hannah's changes would directly be causing these failures and would be good to get your input on the example app changes anyway. Thanks!

imjoehaines:
> LGTM.
> 
> @imjoehaines there are some failing tests (looks like the required Python versions can't be installed??). Please could you take a look at this as I don't think Hannah's changes would directly be causing these failures and would be good to get your input on the example app changes anyway. Thanks!

Fixed in https://github.com/bugsnag/bugsnag-python/pull/330; I'll update this PR when that's merged

imjoehaines (approved):

PR #332, opened 2023-02-03
Extract sub-exceptions from Exception Groups

imjoehaines:
## Goal

Python 3.11 added Exception Groups ([see PEP 654](https://peps.python.org/pep-0654/)) as a way of aggregating several unrelated exceptions into a single Exception-compatible object

bugsnag-python currently supports reporting Exception Groups, but there's no visibility of the sub-exceptions contained within the group. This is problematic as often the group won't contain enough information to debug the problems that caused it — the contained exceptions should be unrelated, so it's difficult to go from the group &rarr; its sub-exceptions

This PR extracts the sub-exceptions from Exception Groups, so that they are reported to bugsnag along with the group. This is very similar to how we report `AggregateExceptions` in the .NET notifier

## Design

In order to avoid truly _gigantic_ lists of errors in the dashboard, we don't recurse into sub-exception groups or their cause/context. The error reporting API also doesn't support tree-shaped hierarchies of errors, so this structure would be difficult to deal with in bugsnag anyway

For example, this exception group:

image

will be reported as:

image

This should still contain enough information to debug the errors and, if not, the exceptions can be reported separately by catching the `ExceptionGroup` itself

### DIFF
--- a/bugsnag/event.py
+++ b/bugsnag/event.py
@@ -17,6 +17,13 @@ from bugsnag.error import Error
 
 __all__ = ('Event',)
 
+if sys.version_info < (3, 11):
+    # an empty tuple can be passed to 'isinstance' safely and will
+    # always return false, so we default to that
+    # this avoids having to wrap any uses of 'BaseExceptionGroup' in a version
+    # check
+    BaseExceptionGroup = ()
+
 
 class Event:
     """
@@ -253,6 +260,21 @@ class Event:
                     self._generate_stacktrace(exception.__traceback__)
                 )
             )
+
+        # unwrap BaseExceptionGroups so that their contained exceptions are
+        # also reported
+        # we don't recurse into nested BaseExceptionGroups or cause/context
+        # here because there's a big risk of that leading to a huge number of
+        # exceptions, which is difficult to reason about
+        if isinstance(exception, BaseExceptionGroup):
+            for sub_exception in exception.exceptions:  # type: ignore
+                error_list.append(
+                    Error(
+                        class_name(sub_exception),
+                        str(sub_exception),
+                        self._generate_stacktrace(sub_exception.__traceback__)
+                    )
+                )
 
         return error_list
 
--- a/tests/fixtures/exception_groups.py
+++ b/tests/fixtures/exception_groups.py
@@ -1,0 +1,27 @@
+# flake8: noqa
+
+# this creates an exception with a very small __traceback__, so it's easier to
+# assert against in tests
+def generate_exception(exception_class, message):
+    try:
+        raise exception_class(message)
+    except Exception as exception:
+        return exception
+
+
+def raise_exception_group_with_no_cause():
+    raise ExceptionGroup(
+        'the message of the group',
+        [
+            generate_exception(Exception, 'exception #1'),
+            generate_exception(ArithmeticError, 'exception #2'),
+            generate_exception(NameError, 'exception #3'),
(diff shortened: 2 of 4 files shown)


tests/test_client.py
```diff
@@ -1493,6 +1493,131 @@ def test_deleting_skip_bugsnag_attr_allows_notify(self):
...
+
+        payload = self.server.received[0]['json_body']
+        exceptions = payload['events'][0]['exceptions']
+
+        # there should be 1 ExceptionGroup with 4 sub-exceptions
+        assert len(exceptions) == 5
+
+        assert exceptions[0] == {
+            'message': 'the message of the group (4 sub-exceptions)',
+            'errorClass': 'builtins.ExceptionGroup',
```
imjoehaines, line 1516: Our `fully_qualified_class_name` is appending `builtins.` here, which is not technically _wrong_ but is unexpected

Fixing this is trivial but I've split this into a separate task as:

1. I'm not sure why this happens for `ExceptionGroup` but not other built-in exceptions
2. `fully_qualified_class_name` is also used when applying `ignore_classes`, so we need to be careful not to change its behaviour in a way that would break ignoring errors

bugsnag/event.py
```diff
@@ -254,6 +261,21 @@ def _generate_error_list(
                 )
             )
 
+        # unwrap BaseExceptionGroups so that their contained exceptions are
+        # also reported
+        # we don't recurse into nested BaseExceptionGroups or cause/context
+        # here because there's a big risk of that leading to a huge number of
+        # exceptions, which is difficult to reason about
+        if isinstance(exception, BaseExceptionGroup):
+            for sub_exception in exception.exceptions:  # type: ignore
```
imjoehaines, line 270: This `type: ignore` comment is because we run `mypy` on Python 3.5 (as well as 3.11) where it has no knowledge of `BaseExceptionGroup`, so complains about using `exception.exceptions`

tests/fixtures/exception_groups.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def raise_exception_group_with_no_cause():
+    raise ExceptionGroup(
+        'the message of the group',
+        [
+            generate_exception(Exception, 'exception #1'),
+            generate_exception(ArithmeticError, 'exception #2'),
+            generate_exception(NameError, 'exception #3'),
+            generate_exception(AssertionError, 'exception #4'),
```
tomlongridge, line 19: Worth adding a nested ExceptionGroup here so you can verify what is, and isn't, reported from it?

tests/fixtures/exception_groups.py
```diff
@@ -0,0 +1,48 @@
...
+    exception_group_with_no_cause = exception_group
+
+
+class MyExceptionGroup(BaseExceptionGroup):
+    pass
+
+
+def raise_base_exception_group_subclass_with_no_cause():
+    raise MyExceptionGroup(
+        'my very easy method just speeds up (n)making exception groups',
```
tomlongridge, line 36: > (n)making

Not sure what you mean by this? Maybe a term I've not heard before...

imjoehaines: `MyExceptionGroup` made me think of this planet mnemonic:

> My Very Easy Method Just Speeds Up Naming Planets

but there are no synonyms for 'making' that start with 'n' ☹️

tomlongridge (approved):

PR #334, opened 2023-02-06
Allow setting the grouping hash when logging

### LINKED ISSUE #253: Unable to set grouping_hash using `logger.<level>()` methods
### Description

There is no way to change the grouping hash value when using logger methods and the BugsnagHandler to send event.

In a general manner, this module API does not make the best use of the standard logging API. That makes it very hard to integrate with existing programs and packages.

**Describe the solution you'd like**

```python
try:
     # Raising code
except:
    logger.exception("This happened", extra={'groupingHash': '<hash value>', ...})
    # Handle the exception.
```

Or to make it even less vendor specific and provide the ability to adopt Bugsnag without modifying a single line of code, while still grouping as one needs, e.g. if the grouping key to use is already available in the log record extras:

```python
import logging
root_logger = logging.getLogger()
# Note that noone should have to write such code by hand (see alternatives section below)
root_logger.addHandler(BugsnagHandler(grouping_hash='<extra-key-name>', ...))

# And later on in the code
try:
     # Raising code
except:
    logger.exception('This happened', extra={'<extra-key-name>': '<hash value>', ...})
    # Handle the exception
```

**Describe alternatives you've considered**

One way to override the `grouping_hash` so far is to use that  keyword parameter of the `bugsnag.notify()` function.
This API is really not great as it imposes to use one that is provider specific in your code when there is no need to.

Another way would be, I believe, to use a [callback](https://github.com/bugsnag/bugsnag-python/blob/master/bugsnag/handlers.py#L80). The problem with this approach is that, as currently implemented, it is not possible to pass callbacks if one wants to configure logging using a standard [configuration file](https://docs.python.org/3/library/logging.config.html). One has to make explicit calls to the `BugsnagHandler.add_callback()` method somewhere in its code. Why would you write code, to do something that should be done through configuration ?

**Additional context**

0HyperCube:
## Goal

Allows changing the grouping hash when using the logger methods and the BugsnagHandler to send event, in order to close #253.

## Changeset

An `extract_grouping_hash` was added to the logging handler to allow the grouping hash to be sent to the `client.notify` function.

## Testing

The `test_logging_grouping_hash` ensures that the specified hash value is sent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Changelog
 
 ### Enhancements
 
-* Allows changing the grouping hash when using the logger methods and
+* Allows changing the grouping hash when using `BugsnagHandler` via the logger methods' `extra` keyword argument
   [#334](https://github.com/bugsnag/bugsnag-python/pull/334)
 
 ### Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Enhancements
+
+* Allows changing the grouping hash when using the logger methods and
+  [#334](https://github.com/bugsnag/bugsnag-python/pull/334)
+
 ## v4.4.0 (2023-02-21)
 
 ### Enhancements
--- a/bugsnag/handlers.py
+++ b/bugsnag/handlers.py
@@ -18,7 +18,8 @@ class BugsnagHandler(logging.Handler, object):
         self.custom_metadata_fields = extra_fields
         self.callbacks = [self.extract_default_metadata,
                           self.extract_custom_metadata,
-                          self.extract_severity]
+                          self.extract_severity,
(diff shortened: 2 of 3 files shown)


johnkiely1:
Hi @0HyperCube,  

Thanks for the PR, we will review when priorities allow.

CHANGELOG.md
```diff
@@ -3,10 +3,17 @@ Changelog
 
 ## TBD
 
+### Enhancements
+
+* Allows changing the grouping hash when using the logger methods and
```
imjoehaines, line 8: Looks like this sentence got cut off. How does this sound?

```suggestion
* Allows changing the grouping hash when using `BugsnagHandler` via the logger methods' `extra` keyword argument
```

0HyperCube: Sounds good.

imjoehaines (approved):

PR #336, opened 2023-02-10
Fix `ExceptionGroup` support when there is also a `__cause__` or `__context__`

imjoehaines:
## Goal

In #332 we now extract sub-exceptions from an `ExceptionGroup`, but this didn't work correctly when the group had a `__cause__` or `__context__`. This is because we re-use the `exception` variable when traversing the cause/context chain, so it would point to the wrong exception object when we checked for an `ExceptionGroup`

We now use the Event's `original_error` instead, which will always point to the reported exception object


tomlongridge (approved):
Wow - that got complicated quickly :)

PR #337, opened 2023-02-15
Test utility improvements

imjoehaines:
## Goal

Fixes some issues I noticed with our test server:

| problem | solution |
|---------|----------|
| `wait_for_request` sleeps for 250ms minimum | sleep in 10ms intervals instead |
| `wait_for_request` _should_ sleep for the entire timeout duration, but the calculation was wrong | fix calculation to `timeout - duration_slept` instead of just sleeping for `duration_slept` |
| the `bugsnag_server` fixture is only available to integration tests | make it available to all tests |
| with the previous problems fixed, tests using `bugsnag_server` take at least 2 seconds to run | only tests for integrations will sleep for the entire timeout, other tests stop waiting as soon as there's a request |

These changes don't affect overall runtime on new Python versions, but do affect older versions quite a lot. CI now takes ~5 minutes to finish, compared to ~3-4 minutes before these changes

This is still pretty acceptable as Python 3.11 will finish in <2 minutes, so you still get feedback just as quickly most of the time — the only problem is when something breaks exclusively on old Python versions


tomlongridge (approved):
Nice :)

PR #340, opened 2023-07-03
Add exception notes as metadata

0HyperCube:
## Goal

Send new `__notes__` field added in https://peps.python.org/pep-0678/

## Changeset

Add a notes tab if the exception has a `__notes__` attribute.

## Testing

New `test_notes` test added to ensure that the notes are added to the payload's metadata.

### DIFF
--- a/bugsnag/event.py
+++ b/bugsnag/event.py
@@ -127,7 +127,7 @@ class Event:
         if hasattr(exception, "__notes__"):
             self.add_tab(
                 "exception notes",
-                dict(enumerate(exception.__notes__))
+                dict(enumerate(exception.__notes__)) # type: ignore # noqa
             )
 
     @property
--- a/bugsnag/event.py
+++ b/bugsnag/event.py
@@ -124,6 +124,9 @@ class Event:
         for name, tab in options.items():
             self.add_tab(name, tab)
 
+        if hasattr(exception, "__notes__"):
+            self.add_tab("notes", dict(enumerate(exception.__notes__)))
+
     @property
     def meta_data(self) -> Dict[str, Dict[str, Any]]:
         warnings.warn('The Event "metadata" property has been replaced ' +
--- a/bugsnag/event.py
+++ b/bugsnag/event.py
@@ -125,7 +125,10 @@ class Event:
             self.add_tab(name, tab)
 
         if hasattr(exception, "__notes__"):
-            self.add_tab("exception notes", dict(enumerate(exception.__notes__)))
+            self.add_tab(
+                "exception notes",
+                dict(enumerate(exception.__notes__))
+            )
 
     @property
     def meta_data(self) -> Dict[str, Dict[str, Any]]:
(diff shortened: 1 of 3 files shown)


tests/test_client.py
```diff
@@ -1056,6 +1056,21 @@ def test_chained_exceptions_with_explicit_cause(self):
             }
         ]
 
+    def test_notes(self):
+        e = Exception("exception")
+        e.add_note("exception note 1")
```
imjoehaines, line 1061: This fails on Python < 3.11 because `add_note` doesn't exist: https://github.com/bugsnag/bugsnag-python/actions/runs/5444670041/jobs/9903374958#step:5:227

You'll need to skip the test on those versions like this: https://github.com/bugsnag/bugsnag-python/blob/2d1db012fb5a87dce94f71d1fa395865d13da389/tests/test_client.py#L572-L575

gareththackeray (approved):

bugsnag/event.py
```diff
@@ -124,6 +124,9 @@ def get_config(key):
         for name, tab in options.items():
             self.add_tab(name, tab)
 
+        if hasattr(exception, "__notes__"):
+            self.add_tab("exception notes", dict(enumerate(exception.__notes__)))
```
imjoehaines, line 128: This fails linting because the line is every so slightly too long

> bugsnag/event.py:128:80: E501 line too long (81 > 79 characters)
>
> https://github.com/bugsnag/bugsnag-python/actions/runs/5452725832/jobs/9920702076?pr=340#step:5:71

```suggestion
            self.add_tab(
                "exception notes",
                dict(enumerate(exception.__notes__))
            )
```

bugsnag/event.py
```diff
@@ -124,6 +124,12 @@ def get_config(key):
         for name, tab in options.items():
             self.add_tab(name, tab)
 
+        if hasattr(exception, "__notes__"):
+            self.add_tab(
+                "exception notes",
+                dict(enumerate(exception.__notes__))
```
imjoehaines, line 130: The lint error for the line being too long covered up a mypy error:

> bugsnag/event.py:130: error: "BaseException" has no attribute "__notes__"

It's fine to just ignore it since we're checking with `hasattr` immediately beforehand:

```suggestion
                dict(enumerate(exception.__notes__)) # type: ignore # noqa
```

tests/test_client.py
```diff
@@ -1056,6 +1056,25 @@ def test_chained_exceptions_with_explicit_cause(self):
...
+        e.add_note("exception note 2")
+        self.client.notify(e)
+        assert self.sent_report_count == 1
+
+        payload = self.server.received[0]['json_body']
+        metadata = payload['events'][0]['metaData']
+        notes = metadata['notes']
```
imjoehaines, line 1072: This is now in `exception notes`

```suggestion
        notes = metadata['exception notes']
```

PR #341, opened 2023-07-03
Fix encoding identical siblings

0HyperCube:
## Goal

Currently when encoding a list or dictionary with identical siblings but no recursion, one of the fields is mistakenly replaced with `[RECURSIVE]`.

## Design

After finishing encoding an object, it is removed from the ignored list.

## Changeset

Identical sibling objects are now properly encoded.

## Testing

New `test_encoding_identical_siblings` test added.

### DIFF
--- a/bugsnag/tornado/__init__.py
+++ b/bugsnag/tornado/__init__.py
@@ -42,7 +42,7 @@ class BugsnagRequestHandler(RequestHandler):
         event.add_tab("request", request_tab)
 
         if bugsnag.configure().send_environment:
-            env = WSGIContainer.environ(self.request)
+            env = WSGIContainer(self.application).environ(self.request)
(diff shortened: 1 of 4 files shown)


bugsnag/tornado/__init__.py
```diff
@@ -42,7 +42,7 @@ def add_tornado_request_to_notification(self, event: bugsnag.Event):
         event.add_tab("request", request_tab)
 
         if bugsnag.configure().send_environment:
-            env = WSGIContainer.environ(self.request)
+            env = WSGIContainer(self.application).environ(self.request)
```
imjoehaines, line 45: Can you back out this change from this PR? I'm happy to merge these PRs with CI failing since we know the cause and can fix it separately before releasing a new version

PR #344, opened 2023-07-04
Accept PathLike as project root

0HyperCube:
## Goal

Currently if a PathLike object is passed in such as PurePath or Path from pathlib, the path will not be set and a warning is produced. The PathLike object should be converted to a string and accepted.

## Changeset

* PathLike objects are now accepted
* Warning message for invalid paths has changed

## Testing

Modification to `test_validate_project_root` to ensure that PathLike objects are accepted and converted to strings.

### DIFF
--- a/bugsnag/configuration.py
+++ b/bugsnag/configuration.py
@@ -8,7 +8,6 @@ import warnings
 import logging
 from threading import Lock
 
-from pathlib import PurePath
 from bugsnag.breadcrumbs import (
     BreadcrumbType,
     Breadcrumb,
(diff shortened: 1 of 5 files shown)


bugsnag/configuration.py
```diff
@@ -8,6 +8,7 @@
 import logging
 from threading import Lock
 
+from os import PathLike
```
imjoehaines, line 11: This was [added in Python 3.6](https://docs.python.org/3/library/os.html#os.PathLike) so the import fails in Python 3.5, which we still support

I think the best we can do on 3.5 is to accept `Union[str, PurePath]` instead, since all the builtin `Path` objects inherit from `PurePath`

We do things like that elsewhere by catching the `ImportError`, e.g.:

https://github.com/bugsnag/bugsnag-python/blob/cc6aa2027b3c152b35b984638f9c749f25e7d39f/bugsnag/sessiontracker.py#L8-L14

0HyperCube: I've switched it to use PurePath instead of PathLike.

imjoehaines (approved):

PR #347, opened 2023-07-20
Add FeatureFlag class

imjoehaines:
## Goal

Add `FeatureFlag` class that stores a `name` and an optional `variant`, e.g.:

```python
flag1 = FeatureFlag('name of the flag')
flag2 = FeatureFlag('another name', 'variant')
```


tomlongridge (approved):
LGTM 👍

PR #355, opened 2023-08-29
Prevent feature flags from leaking between clients

imjoehaines:
## Goal

Currently feature flags are stored in the same place across all Client instances, meaning every client shares the same list of feature flags (though they are stored in a ContextVar/thread local so don't leak between async contexts/threads)

This PR introduces a new `ContextLocalState` API that allows storing things separately per-client and per-context. This works by storing a `WeakKeyDictionary` in a `ContextVar` with `Client` instances as keys and a dict as values. The `WeakKeyDictionary` ensures that that when Clients are garbage collected their contextual state is collected too

In the future this can be extended to cover other contextual data (breadcrumbs, sessions etc&hellip;) but that would require a lot of refactoring and could be considered a BC break so will be implemented in a future PR for the next major version


tests/test_context.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.mark.skipif(
+    sys.version_info < (3, 7),
+    reason="requires ContextVar support (Python 3.7 or higher)"
+)
+def test_state_is_stored_separately_per_async_context():
+    client1 = Client()
+    client2 = Client()
+
+    assert client1.feature_flags == []
+    assert client2.feature_flags == []
```
tomlongridge, line 92: Is there a reason that you're not testing with some initial flags in the clients (as per preceding test)?

imjoehaines: nope! it works fine with initial flags

tomlongridge (approved):

PR #373, opened 2024-02-07
fix: utcnow() warning with python 3.12

Andarius:
Fix warning for python 3.12 on utcnow()

## Goal

Remove the following deprecation warning:

DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC)

## Design

Since datetule.UTC does not exists in previous versions, 

## Changeset

Replaced utcnow()  with `dt.datetime.now(timezone.utc).replace(tzinfo=None)`

### DIFF
--- a/bugsnag/utils.py
+++ b/bugsnag/utils.py
@@ -6,7 +6,7 @@ from typing import AnyStr, Tuple, Optional
 import warnings
 import copy
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from urllib.parse import urlparse, urlunsplit, parse_qs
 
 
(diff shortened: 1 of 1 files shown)


fabiob:
Why was this closed? Can I help somehow?

thimi0412:
I have the same problem.

mclack:
Hi all,

We have opened our own PR to replace `utcnow` calls: https://github.com/bugsnag/bugsnag-python/pull/394

While I don't have a clear ETA on when this will be released, we will make sure to put this through our full review process when priorities allow. Please feel free to keep an eye on the linked PR for updates.

PR #376, opened 2024-03-27
Track in flight requests

imjoehaines:
## Goal

This PR adds a `RequestTracker` class that can be used to (surprise surprise) track requests:

```python
request_tracker = RequestTracker()
mark_request_complete = request_tracker.new_request()
request_tracker.has_in_flight_requests() # => True

# ...make the request...

mark_request_complete()
request_tracker.has_in_flight_requests() # => False
```

The `Client` and `SessionTracker` use this, along with post-delivery callbacks (#375), to keep track of when there are in-flight event or session requests

### DIFF
--- a/bugsnag/client.py
+++ b/bugsnag/client.py
@@ -19,6 +19,7 @@ from bugsnag.handlers import BugsnagHandler
 from bugsnag.sessiontracker import SessionTracker
 from bugsnag.utils import to_rfc3339
 from bugsnag.context import ContextLocalState
+from bugsnag.request_tracker import RequestTracker
 
 __all__ = ('Client',)
 
@@ -36,6 +37,7 @@ class Client:
         self.session_tracker = SessionTracker(self.configuration)
         self.configuration.configure(**kwargs)
         self._context = ContextLocalState(self)
+        self._request_tracker = RequestTracker()
 
         if install_sys_hook:
             self.install_sys_hook()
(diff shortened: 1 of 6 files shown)


bugsnag/client.py
```diff
@@ -192,16 +189,30 @@ def send_payload():
...
+                        options
+                    )
                 except Exception as e:
                     self.configuration.logger.exception(
                         'Notifying Bugsnag failed %s',
                         e
                     )
 
+                    # ensure this request is not still marked as in-flight
+                    post_delivery_callback()
```
Cawllec, line 214: Given the standard delivery has finally statements to call this baked in, do we still need to have this here?  I guess in the case of an exception before we get to that point?

tests/test_client.py
```diff
@@ -1688,6 +1693,62 @@ def on_error(event):
...
+        configuration = Configuration()
+        configuration.configure(delivery=delivery, api_key='abc')
+
+        client = Client(configuration)
+        assert not client._request_tracker.has_in_flight_requests()
+
+        client.notify(Exception('Oh no'))
+        assert client._request_tracker.has_in_flight_requests()
+
+        delivery.flush_request_queue()
```
Cawllec, line 1707: I'm assuming this is synchronous and won't introduce a race condition with the following check?

bugsnag/client.py
imjoehaines: Yeah, it's for exactly this:

> in the case of an exception before we get to that point?

tests/test_client.py
imjoehaines: Yeah, [`QueueingDelivery` came from #375](https://github.com/bugsnag/bugsnag-python/pull/375/files#diff-9e08e8c7769db7b58089fd7659d75ca96df40bb7b5d50299d3a30abd54da9ad6R182-R193) and is entirely synchronous

Cawllec (approved):

PR #381, opened 2024-04-12
Fix unhandled exceptions in lambdas being marked as handled

imjoehaines:


Cawllec (approved):

imjoehaines:
Merging as CI failure is unrelated (fixed in https://github.com/bugsnag/bugsnag-python/pull/382)

PR #393, opened 2024-11-01
update utcnow to datetime.now(timezone.utc)

clr182:
## Goal

DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC)

## Changeset

Changed utcnow in bugsnag/utils.py to `datetime.now(timezone.utc)`


clr182:
https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources 
Maybe we should use setuptools rather than explicitly setting pkg-resources. Setuptools seems to include pkg-resources.

PR #399, opened 2025-06-20
Set default endpoints based on API key

twometresteve:
## Goal

Set default endpoints based on API key. If no endpoint is configured, payloads should be sent to `*.insighthub.smartbear.com` if the API key starts with `00000`.

## Design

The logic that defaults the endpoints based on the API key depends on `configure` not being called before it is called with the API key.  Although our integration guide directs people to use `bugsnag.configure`(for a single Bugsnag client), it is also possible to use multiple `Client`.  In singleton mode (using `legacy.py`) `Client` is instantiated and calls `configure` without any arguments - breaking that dependency.

As a solution to that, avoiding breaking existing uses of `Client`, I've added a `configure` parameter to the `Client` constructor, defaulting to true but allowing the "legacy" mode to stop it being called (with no actual arguments).

## Testing

Unit tests added and the existing tests also caught an issue in development.  I've also run some manual tests locally with an example app, verifying that the requested host is as expected for different API keys.  That exercise highlighted the issue with single/multiple clients.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@ Changelog
 
 ### Enhancements
 
-* Remove depricated `datetime.utcnow()` method call from utils class
+* Remove deprecated `datetime.utcnow()` method call from utils class
   [#394](https://github.com/bugsnag/bugsnag-python/pull/394).
+* Set default endpoints based on API key
+  [#399](https://github.com/bugsnag/bugsnag-php/pull/399)
 
 ## v4.7.1 (2024-05-22)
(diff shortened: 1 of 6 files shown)


tomlongridge (approved):
As discussed IRL, the semi-internal new parameter seems the least intrusive or potential for breaking option. It's tempting to create a new client when `bugsnag.configure` is called, but it's potential that users might be relying on some side-effect of having the same client instance, so I'm happy to avoid that uncertainty with this approach.

PR #410, opened 2026-03-27
Pawan/plat 14675 flask lambda

SB-PawanN:
## Goal

Problem: Prevent legacy module/package configuration (import-time snapshot) from becoming stale and diverging from the live Configuration.

## Design

Approach: Defer synchronization to [configure()] and perform call-time, best-effort, non-throwing synchronization of legacy/module/package references.

## Changeset

* [legacy.py]: Call-time sync of [configuration], [default_client.configuration], and [logger]; update [bugsnag] package attrs; remove unused global.
* [test_legacy_configuration.py]: Add regression test asserting [bugsnag.configuration is legacy.configuration] (and that [logger] reflects the live instance).
* [tox.ini]: Pin starlette<0.28 and httpx<0.28 to stabilize ASGI tests.

## Testing

Unit: Added [test_legacy_configuration.py].
Local CI: Ran tox -e py310-asgi, tox -e py313-lint, tox -e py314-lint — all passed locally.
Notes: Lint fix addressed a long line (E501) and an unused global; no test-only shims are included in the changeset.

### DIFF
--- a/bugsnag/legacy.py
+++ b/bugsnag/legacy.py
@@ -1,6 +1,7 @@
 from typing import Dict, Any, Tuple, Type, Optional, Union, List, Callable
 import types
 import sys
+import warnings
 
 from bugsnag.breadcrumbs import BreadcrumbType, OnBreadcrumbCallback
 from bugsnag.feature_flags import FeatureFlag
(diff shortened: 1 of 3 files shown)


SB-PawanN:
Closing this as it cannot be reproduced.

